PR #2, opened 2016-11-22
Changed failure handling of the agent overlay module.

asridharan:
The PR contains the following changes:
a) Introduced a timeout for all invocations to the docker daemon.
b) Also introduced an `onDiscard` callback to `Subprocess` invocations, to kill the subprocess when the callee discards the future.
c) Have introduced new fields in the agent configuration in order to allow the operator to specify the location of the `ipset`, `iptabels` and `docker` binaries. This also  helps us in emulating failures of these binaries during unit-tests.
d) Introduced a temp directory for CNI configuration. Since, now we are allowing the overlay module to progress even if the overlay configuration has not been successful, we need to ensure that the CNI config is installed only after successful overlay configuration. To achieve this we introduce a temp directory where the CNI configuration is installed. The CNI configuration is moved from the temp location to the CNI config directory once the overlay configuration is successful.

### DIFF
--- a/common/shell.hpp
+++ b/common/shell.hpp
@@ -3,6 +3,8 @@
 
 #include <vector>
 
+#include <stout/os/kill.hpp>
+
 #include <process/collect.hpp>
 #include <process/io.hpp>
 #include <process/future.hpp>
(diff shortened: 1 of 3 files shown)


asridharan:
Deprecating this pull request in favor of PR#4

PR #4, opened 2016-12-01
Restructured agent overlay-module

asridharan:
* Exposed the `agent::ManagerProcess` via agent.hpp
* Modified unit-tests to use `agent::ManagerProcess` in their test cases instead of the overlay agent module directly.


asridharan:
Discarding this review for #8

PR #5, opened 2016-12-07
Fixed containerLogger child environment setup.

tillt:


journald/lib_journald.cpp
```diff
@@ -64,14 +64,22 @@ class JournaldContainerLoggerProcess :
...
-    // will conflict with the already-running agent.
-    std::map<std::string, std::string> environment = os::environment();
-    environment.erase("LIBPROCESS_PORT");
-    environment.erase("LIBPROCESS_ADVERTISE_PORT");
+    // Prepare the environment for the container logger subprocess.
+    // We inherit agent environment variables except for those
+    // LIBPROCESS or MESOS prefixed environment variables. See MESOS-6747.
+    map<string, string> environment;
+
+    foreachpair (const string& key, const string& value, os::environment()) {
```
kaysoky: This file doesn't import `std::string` nor `std::map` (and shouldn't).  So you'll need to qualify these types.

PR #6, opened 2016-12-09
Modified the re-registration logic to allow change to agent config.

asridharan:
Added an `UpdateAgent` `operation` to the overlay replicated log to
update `AgentInfo` for agents whose overlay configuration might need
to be changed due to a change in the agent's network config. The
change in agent's network config could be the removal or addition of a
`docker_bridge` or a `mesos_bridge` and the addition or removal of an
`allocated_subnet`.

### DIFF
--- a/overlay/agent.cpp
+++ b/overlay/agent.cpp
@@ -260,6 +260,26 @@ protected:
 
       overlays[name] = overlay;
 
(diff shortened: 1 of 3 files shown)


asridharan:
We don't want to implement the re-conciliation logic at this piont so going to discard this PR at the moment.

PR #7, opened 2016-12-09
Modified Mesos and Docker configuration in the Agent.

asridharan:
Instead of relying on the information sent by the Master, we create
Mesos and Docker networks based on the configuration given by the
operator.

### DIFF
--- a/overlay/agent.cpp
+++ b/overlay/agent.cpp
@@ -527,8 +527,8 @@ protected:
     // The below command is a script consisting of three commands:
     // <set ipset> && <check iptables rule exists> ||
     // <insert iptables rule>
-    if (!overlays[name].has_mesos_bridge() &&
-        !overlays[name].has_docker_bridge()) {
+    if (!networkConfig.mesos_bridge() &&
+        !networkConfig.docker_bridge()) {
       return overlaySuccess("");
     } else {
       const string overlaySubnet = overlays[name].info().subnet();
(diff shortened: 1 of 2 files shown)


overlay/agent.cpp
```diff
@@ -527,8 +527,8 @@ class ManagerProcess : public ProtobufProcess<ManagerProcess>
     // The below command is a script consisting of three commands:
     // <set ipset> && <check iptables rule exists> ||
     // <insert iptables rule>
-    if (!overlays[name].has_mesos_bridge() &&
-        !overlays[name].has_docker_bridge()) {
+    if (!networkConfig.mesos_bridge() &&
+        !networkConfig.docker_bridge()) {
```
jieyu: print a WARNING if they are different.

kaysoky:
The branch name `DCOS-11982-HACK` gives me great confidence 😛 .

Looks reasonable to me.

asridharan:
@jieyu @kaysoky added a unit-test to test the behavior of the agent when its network config is changed in between agent restarts.

asridharan:
Merged this.

asridharan:
commit 3d185b433c999a448f5648876b498cc2537ba554
Author: Avinash Sridharan <avinash@mesosphere.com>
Date:   Fri Dec 9 22:16:24 2016 +0000

    Modified Mesos and Docker configuration in the Agent.

    Instead of relying on the information sent by the Master, we create
    Mesos and Docker networks based on the configuration given by the
    operator.

    With this change the semantics of Master Agent interaction change to
    the overlay configuration being sent to the agent being treated as a
    request that the agent will try to comply to as long as the operator
    allows the agent to apply the configuration requested by the Master.

commit 1073246c3dd9a9098ad79918048543a79ef22520
Author: Avinash Sridharan <avinash@mesosphere.com>
Date:   Sat Dec 10 01:24:27 2016 +0000

    Added unit-test to verify behavior of `AgentNetworkConfig`.

    The test verifies that when agent is restarted with an
    `AgentNetworkConfig` different than the one with which it has
    registered with the Master, though the Master might send an overlay
    configuration that is contrary to what the `AgentNetworkConfig`
    demands (E.g., `AgentNetworkConfig` has `mesos_bridge` and
    `docker_bridge` disabled but Master sends an overlay configuration
    with both enabled) the agent always honors the configuration specified
    in `AgentNetworkConfig` over what the Master sends.

PR #9, opened 2017-01-06
Combined Logrotate logger functionality into journald logger.

kaysoky:
This provides a backwards compatibility mode for the container logger.
In case the sandbox logs are still necessary, this adds a "destination_type" option which can pipe logs to both journald and the sandbox.

### DIFF
--- a/journald/journald.cpp
+++ b/journald/journald.cpp
@@ -133,7 +133,7 @@ public:
           return Nothing();
         }
 
-        if (flags.destination_type == "logrotate" ||
+        if (flags.destination_type == "journald" ||
             flags.destination_type == "journald+logrotate") {
           // Write the bytes to journald.
           Try<Nothing> result = write_journald(readSize);
@@ -143,7 +143,7 @@ public:
           }
         }
 
-        if (flags.destination_type == "journald" ||
+        if (flags.destination_type == "logrotate" ||
             flags.destination_type == "journald+logrotate") {
           // Write the bytes to sandbox, with log rotation.
           Try<Nothing> result = write_logrotate(readSize);
--- a/journald/journald.cpp
+++ b/journald/journald.cpp
@@ -7,6 +7,7 @@
 
 #include <process/dispatch.hpp>
 #include <process/future.hpp>
+#include <process/id.hpp>
 #include <process/io.hpp>
 #include <process/process.hpp>
 
@@ -17,6 +18,8 @@
 #include <stout/try.hpp>
 
 #include <stout/os/pagesize.hpp>
+#include <stout/os/shell.hpp>
+#include <stout/os/su.hpp>
 
 #include "journald.hpp"
 
@@ -29,7 +32,8 @@ class JournaldLoggerProcess : public Process<JournaldLoggerProcess>
 {
 public:
   JournaldLoggerProcess(const Flags& _flags)
-    : flags(_flags)
+    : ProcessBase(process::ID::generate("journald-logger")),
+      flags(_flags)
   {
     // Prepare a buffer for reading from the `incoming` pipe.
     length = os::pagesize();
@@ -54,6 +58,10 @@ public:
       delete[] entries;
       entries = NULL;
     }
+
+    if (leading.isSome()) {
+      os::close(leading.get());
+    }
   }
 
   // Prepares and starts the loop which reads from stdin and writes to journald.
@@ -73,6 +81,27 @@ public:
       entries[i].iov_len = entry.length();
       entries[i].iov_base = new char[entry.length() + 1];
       std::strcpy((char*) entries[i].iov_base, entry.c_str());
+    }
+
+    if (flags.destination_type != "journald") {
+      // Populate the `logrotate` configuration file.
+      // See `Flags::logrotate_options` for the format.
+      //
+      // NOTE: We specify a size of `--max_size - length` because `logrotate`
+      // has slightly different size semantics.  `logrotate` will rotate when the
+      // max size is *exceeded*.  We rotate to keep files *under* the max size.
+      const std::string config =
+        "\"" + flags.log_filename.get() + "\" {\n" +
+        flags.logrotate_options.getOrElse("") + "\n" +
+        "size " + stringify(flags.max_size.bytes() - length) + "\n" +
+        "}";
+
+      Try<Nothing> result = os::write(
+          flags.log_filename.get() + CONF_SUFFIX, config);
+
+      if (result.isError()) {
+        return Failure("Failed to write configuration file: " + result.error());
+      }
     }
 
     // NOTE: This is a prerequisuite for `io::read`.
@@ -100,11 +129,22 @@ public:
           return Nothing();
         }
 
-        // Write the bytes to journald.
-        Try<Nothing> result = write(readSize);
-        if (result.isError()) {
-          promise.fail("Failed to write: " + result.error());
-          return Nothing();
+        if (flags.destination_type != "sandbox") {
+          // Write the bytes to journald.
+          Try<Nothing> result = write_journald(readSize);
+          if (result.isError()) {
+            promise.fail("Failed to write: " + result.error());
+            return Nothing();
+          }
+        }
+
+        if (flags.destination_type != "journald") {
+          // Write the bytes to sandbox, with log rotation.
+          Try<Nothing> result = write_logrotate(readSize);
+          if (result.isError()) {
+            promise.fail("Failed to write: " + result.error());
+            return Nothing();
+          }
         }
 
         // Use `dispatch` to limit the size of the call stack.
@@ -116,7 +156,7 @@ public:
 
   // Writes the buffer from stdin to the journald.
   // Any `flags.labels` will be prepended to each line.
-  Try<Nothing> write(size_t readSize)
+  Try<Nothing> write_journald(size_t readSize)
   {
     // We may be reading more than one log line at once,
     // but we need to add labels for each line.
@@ -140,12 +180,83 @@ public:
     return Nothing();
   }
 
+
+  // Writes the buffer from stdin to the leading log file.
+  // When the number of written bytes exceeds `--max_size`, the leading
+  // log file is rotated.  When the number of log files exceed `--max_files`,
+  // the oldest log file is deleted.
+  Try<Nothing> write_logrotate(size_t readSize)
+  {
+    // Rotate the log file if it will grow beyond the `--max_size`.
+    if (bytesWritten + readSize > flags.max_size.bytes()) {
+      rotate();
+    }
+
+    // If the leading log file is not open, open it.
+    // NOTE: We open the file in append-mode as `logrotate` may sometimes fail.
+    if (leading.isNone()) {
+      Try<int> open = os::open(
+          flags.log_filename.get(),
+          O_WRONLY | O_CREAT | O_APPEND | O_CLOEXEC,
+          S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH);
+
+      if (open.isError()) {
+        return Error(
+            "Failed to open '" + flags.log_filename.get() +
+            "': " + open.error());
+      }
+
+      leading = open.get();
+    }
+
+    // Write from stdin to `leading`.
+    // NOTE: We do not exit on error here since we are prioritizing
+    // clearing the STDIN pipe (which would otherwise potentially block
+    // the container on write) over log fidelity.
+    Try<Nothing> result =
+      os::write(leading.get(), std::string(buffer, readSize));
+
+    if (result.isError()) {
+      std::cerr << "Failed to write: " << result.error() << std::endl;
+    }
+
+    bytesWritten += readSize;
+
+    return Nothing();
+  }
+
+  // Calls `logrotate` on the leading log file and resets the `bytesWritten`.
+  void rotate()
+  {
+    if (leading.isSome()) {
+      os::close(leading.get());
+      leading = None();
+    }
+
+    // Call `logrotate` to move around the files.
+    // NOTE: If `logrotate` fails for whatever reason, we will ignore
+    // the error and continue logging.  In case the leading log file
+    // is not renamed, we will continue appending to the existing
+    // leading log file.
+    os::shell(
+        flags.logrotate_path +
+        " --state \"" + flags.log_filename.get() + STATE_SUFFIX + "\" \"" +
+        flags.log_filename.get() + CONF_SUFFIX + "\"");
+
+    // Reset the number of bytes written.
+    bytesWritten = 0;
+  }
+
 private:
   Flags flags;
 
   // For reading from stdin.
   char* buffer;
   size_t length;
+
+  // For writing and rotating the leading log file.
+  Option<int> leading;
+  size_t bytesWritten;
 
   // Used as arguments for `sd_journal_sendv`.
   // This contains one more entry than the number of `--labels`.
@@ -174,6 +285,16 @@ int main(int argc, char** argv)
   // Log any flag warnings.
   foreach (const flags::Warning& warning, load->warnings) {
     LOG(WARNING) << warning.message;
+  }
+
+  // If the `--user` flag is set, change the UID of this process to that user.
+  if (flags.user.isSome()) {
+    Try<Nothing> result = os::su(flags.user.get());
+
+    if (result.isError()) {
+      EXIT(EXIT_FAILURE)
+        << ErrnoError("Failed to switch user for journald logger").message;
+    }
   }
 
   // Asynchronously control the flow and size of logs.
--- a/journald/journald.cpp
+++ b/journald/journald.cpp
@@ -64,7 +64,8 @@ public:
     }
   }
 
-  // Prepares and starts the loop which reads from stdin and writes to journald.
+  // Prepares and starts the loop which reads from stdin and writes to
+  // journald or the sandbox, depending on the input flags.
   Future<Nothing> run()
   {
     // Pre-populate the `iovec` with the constant labels.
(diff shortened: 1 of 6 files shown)


journald/journald.cpp
```diff
@@ -100,11 +129,22 @@ class JournaldLoggerProcess : public Process<JournaldLoggerProcess>
           return Nothing();
         }
 
-        // Write the bytes to journald.
-        Try<Nothing> result = write(readSize);
-        if (result.isError()) {
-          promise.fail("Failed to write: " + result.error());
-          return Nothing();
+        if (flags.destination_type != "sandbox") {
```
adam-mesos: `if (flags.destination_type == "journald" || flags.destination_type == "sandbox_and_journald") {`
Or maybe `flags.destination_types` should take a list of strategies to follow, and you just have to check if the list contains "journald" or "sandbox"

journald/journald.cpp
```diff
@@ -100,11 +129,22 @@ class JournaldLoggerProcess : public Process<JournaldLoggerProcess>
...
+          // Write the bytes to journald.
+          Try<Nothing> result = write_journald(readSize);
+          if (result.isError()) {
+            promise.fail("Failed to write: " + result.error());
+            return Nothing();
+          }
+        }
+
+        if (flags.destination_type != "journald") {
```
adam-mesos: Test for the positive, not the negative, please. "sandbox" || "sandbox_and_journald"

journald/journald.cpp
```diff
@@ -100,11 +129,22 @@ class JournaldLoggerProcess : public Process<JournaldLoggerProcess>
...
+          // Write the bytes to sandbox, with log rotation.
```
adam-mesos: Just wondering why this differs from the OSS comment, "Do log rotation (if necessary) and write the bytes to the leading log file."

journald/journald.cpp
```diff
@@ -176,6 +287,16 @@ int main(int argc, char** argv)
     LOG(WARNING) << warning.message;
   }
 
+  // If the `--user` flag is set, change the UID of this process to that user.
+  if (flags.user.isSome()) {
+    Try<Nothing> result = os::su(flags.user.get());
+
+    if (result.isError()) {
+      EXIT(EXIT_FAILURE)
+        << ErrnoError("Failed to switch user for journald logger").message;
```
adam-mesos: Is this linux user used for for logrotate, journald, or both? Looks to me like that's who calls the logrotate binary via the `shell` calls. But maybe the journald `subprocess` also uses this?

journald/journald.hpp
```diff
@@ -13,12 +13,17 @@
...
+#include <stout/os/shell.hpp>
+
 
 namespace mesos {
 namespace journald {
 namespace logger {
 
 const std::string NAME = "mesos-journald-logger";
+const std::string CONF_SUFFIX = ".logrotate.conf";
+const std::string STATE_SUFFIX = ".logrotate.state";
```
adam-mesos: LOGROTATE_CONF_SUFFIX and LOGROTATE_STATE_SUFFIX, since these are only meaningful if your logger destination type includes logrotated sandbox

journald/journald.hpp
```diff
@@ -32,6 +37,19 @@ struct Flags : public virtual flags::FlagsBase
       "before it is written to journald.  See '--labels'."
       "\n");
 
+    add(&Flags::destination_type,
+        "destination_type",
+        "Determines where logs should be piped.\n"
+        "Valid destinations include: 'journald', 'sandbox', or 'both'.",
```
adam-mesos: 1. "both" is vague and not future-proof in case we add more types later (syslog, sandbox-without-logrotate); please either rename it to something like "journald_and_sandbox" or turn it into a `,`-delimited list like "journald,sandbox" and we can check for set.contains() when choosing when to apply a strategy's behavior.
2. "sandbox" seems like a misnomer, since you actually copied the `LogrotateLoggerProcess`, and there's a separate `SandboxContainerLoggerProcess` that doesn't do log-rotation.

journald/journald.hpp
```diff
@@ -32,6 +37,19 @@ struct Flags : public virtual flags::FlagsBase
...
+        "journald",
```
adam-mesos: Why is the module default "journald", while the proposed dcos default is "both"? Shouldn't these be consistent, or just set/defaulted in a single place?

journald/journald.hpp
```diff
@@ -65,12 +83,86 @@ struct Flags : public virtual flags::FlagsBase
...
+    add(&Flags::logrotate_options,
+        "logrotate_options",
+        "Additional config options to pass into 'logrotate'.\n"
+        "This string will be inserted into a 'logrotate' configuration file.\n"
+        "i.e.\n"
+        "  /path/to/<log_filename> {\n"
+        "    <logrotate_options>\n"
+        "    size <max_size>\n"
+        "  }\n"
+        "NOTE: The 'size' option will be overriden by this command.");
```
adam-mesos: s/overriden/overridden/g here, Apache Mesos, everywhere

journald/lib_journald.cpp
```diff
@@ -93,6 +93,49 @@ class JournaldContainerLoggerProcess :
     environment["LIBPROCESS_NUM_WORKER_THREADS"] =
       stringify(flags.libprocess_num_worker_threads);
 
+    // Copy the global rotation flags.
+    // These will act as the defaults in case the executor environment
+    // overrides a subset of them.
+    LoggerFlags overridenFlags;
```
adam-mesos: overriddenFlags

journald/lib_journald.cpp
```diff
@@ -93,6 +93,49 @@ class JournaldContainerLoggerProcess :
     environment["LIBPROCESS_NUM_WORKER_THREADS"] =
       stringify(flags.libprocess_num_worker_threads);
 
+    // Copy the global rotation flags.
```
adam-mesos: s/rotation flags/logger flags/

journald/lib_journald.cpp
```diff
@@ -93,6 +93,49 @@ class JournaldContainerLoggerProcess :
...
+    // These will act as the defaults in case the executor environment
+    // overrides a subset of them.
+    LoggerFlags overridenFlags;
+    overridenFlags.destination_type = flags.destination_type;
+    overridenFlags.max_stdout_size = flags.max_stdout_size;
```
adam-mesos: s/max_stdout_size/logrotate_max_stdout_size/
ditto for stderr_size

journald/lib_journald.hpp
```diff
@@ -18,10 +18,105 @@ namespace journald {
...
+
+  Bytes max_stdout_size;
+  Option<std::string> logrotate_stdout_options;
+
+  Bytes max_stderr_size;
+  Option<std::string> logrotate_stderr_options;
+};
+
+
+struct Flags : public virtual LoggerFlags
```
adam-mesos: How is it that the generically name `Flags` is a subclass of the more specifically named `LoggerFlags`?
Why does Flags have 4 extra fields, at least one for logrotate, others for both strategies? Why do you need two Flags classes? Please name them appropriately, or merge them.
And even `LoggerFlags` is not specific enough, since it only covers ContainerLogging, and not master/agent logs.

journald/lib_journald.hpp
```diff
@@ -18,10 +18,105 @@ namespace journald {
...
 {
   Flags()
   {
+    add(&Flags::environment_variable_prefix,
+        "environment_variable_prefix",
+        "Prefix for environment variables meant to modify the behavior of\n"
+        "the logrotate logger for the specific executor being launched.\n"
+        "The logger will look for four prefixed environment variables in the\n"
```
adam-mesos: s/four/five/ (or more?)

journald/lib_journald.hpp
```diff
@@ -18,10 +18,105 @@ namespace journald {
...
+        "'ExecutorInfo's 'CommandInfo's 'Environment':\n"
+        "  * DESTINATION_TYPE\n"
+        "  * MAX_STDOUT_SIZE\n"
+        "  * LOGROTATE_STDOUT_OPTIONS\n"
+        "  * MAX_STDERR_SIZE\n"
+        "  * LOGROTATE_STDERR_OPTIONS\n"
+        "If present, these variables will overwrite the global values set\n"
```
adam-mesos: s/overwrite/override/ since they do not modify the global values themselves, the just override the global values for a particular container

journald/lib_journald.hpp
```diff
@@ -68,6 +185,12 @@ struct Flags : public virtual flags::FlagsBase
 // newlines) and adds identifying labels to each log line; such
 // as the FrameworkID, ExecutorID, ContainerID, and labels inside
 // ExecutorInfo.
```
adam-mesos: This part isn't entirely accurate given your 3-mode update.

journald/lib_journald.hpp
```diff
@@ -18,10 +18,105 @@ namespace journald {
 
 class JournaldContainerLoggerProcess;
 
-struct Flags : public virtual flags::FlagsBase
+
+// These flags are loaded twice: once when the `ContainerLogger` module
+// is created and each time before launching executors. The flags loaded
+// at module creation act as global default values, whereas flags loaded
+// prior to executors can override the global values.
+struct LoggerFlags : public virtual flags::FlagsBase
```
kaysoky: I'd argue that the fully qualified name (`mesos::journald::LoggerFlags`) is sufficient to say that this doesn't apply to all logs.

journald/lib_journald.hpp
```diff
@@ -18,10 +18,105 @@ namespace journald {
...
+{
+  LoggerFlags()
+  {
+    add(&LoggerFlags::destination_type,
+        "destination_type",
+        "Determines where logs should be piped.\n"
+        "Valid destinations include: 'journald', 'sandbox', or 'both'.",
+        "journald",
```
kaysoky: I used different defaults because this module's original intent was for the `journald` mode.  Whereas the DC/OS PR explicitly wanted to have both options for the moment.

journald/lib_journald.hpp
kaysoky: The subclassing is explained in the comment above `LoggerFlags`:
```
// These flags are loaded twice: once when the `ContainerLogger` module
// is created and each time before launching executors. The flags loaded
// at module creation act as global default values, whereas flags loaded
// prior to executors can override the global values.
```

journald/lib_journald.hpp
kaysoky: Definitely don't want executors changing the `companion_dir` (which will just cause their executor to fail).  

`libprocess_num_worker_threads` is more of an optimization that is more important for the operator (who might be concerned about running out of threads) than the executor, as the executor should be isolated from other processes on the node.

journald/lib_journald.hpp
```diff
@@ -41,6 +136,25 @@ struct Flags : public virtual flags::FlagsBase
...
+    add(&Flags::logrotate_path,
+        "logrotate_path",
+        "If specified, the logrotate container logger will use the specified\n"
+        "'logrotate' instead of the system's 'logrotate'.",
+        "logrotate",
+        [](const std::string& value) -> Option<Error> {
+          // Check if `logrotate` exists via the help command.
+          // TODO(josephw): Consider a more comprehensive check.
+          Try<std::string> helpCommand =
+            os::shell(value + " --help > /dev/null");
```
kaysoky: Nope, these flags can only be changed by the operator (in `/opt/mesosphere/etc/...`)

journald/journald.cpp
```diff
@@ -29,7 +32,8 @@ class JournaldLoggerProcess : public Process<JournaldLoggerProcess>
 {
 public:
   JournaldLoggerProcess(const Flags& _flags)
-    : flags(_flags)
+    : ProcessBase(process::ID::generate("journald-logger")),
```
kaysoky: Composing `ContainerLoggers` would be clunky.  The end result of composition would be twice the number of logger subprocesses (i.e. 2 logrotate loggers, 2 journald loggers) and a bit of funky action in a `ComposingContainerLogger`, which would need to load modules (as a module itself).

journald/journald.cpp
```diff
@@ -75,6 +83,27 @@ class JournaldLoggerProcess : public Process<JournaldLoggerProcess>
       std::strcpy((char*) entries[i].iov_base, entry.c_str());
```
kaysoky: I left these as-is because this initialization happens only once at the start.  Within the loop, there is only one copy (which copies a pointer), and this only runs in the `journald` mode

journald/journald.cpp
kaysoky: Internal consistency, I suppose.  The matching comment above is `// Write the bytes to journald.`

journald/journald.cpp
```diff
@@ -75,6 +83,27 @@ class JournaldLoggerProcess : public Process<JournaldLoggerProcess>
...
     }
 
+    if (flags.destination_type != "journald") {
```
kaysoky: Just for the record, adding a third option sounds excessive (7 more configurations to test/support!).  

If not for backwards compatibility, we'd only have two options (`sandbox` or `journald`) (not `both`).  And this would be achieved entirely via configuration.  Future destination types should not require changes to this module.

journald/journald.cpp
kaysoky: This is more important for the `logrotate` portion, as logging will break if you don't switch the user appropriately.

I don't think journald has any restrictions on what user uses their APIs.

journald/journald.hpp
```diff
@@ -65,12 +83,86 @@ struct Flags : public virtual flags::FlagsBase
...
+    add(&Flags::logrotate_path,
+        "logrotate_path",
+        "If specified, this command will use the specified\n"
+        "'logrotate' instead of the system's 'logrotate'.",
+        "logrotate",
+        [](const std::string& value) -> Option<Error> {
+          // Check if `logrotate` exists via the help command.
+          // TODO(josephw): Consider a more comprehensive check.
+          Try<std::string> helpCommand =
+            os::shell(value + " --help > /dev/null");
```
kaysoky: Nope.  The environment is inherited from the agent (with some subtractions).  The user can't change this option.

journald/journald.hpp
kaysoky: 1. The intention is to not add anymore types to this logger, but to add any future loggers as config changes.

2. 👍

journald/journald.hpp
kaysoky: Answered earlier.

journald/journald.hpp
```diff
@@ -65,12 +83,86 @@ struct Flags : public virtual flags::FlagsBase
...
+            return Error(
+                "Failed to check logrotate: " + helpCommand.error());
+          }
+
+          return None();
+        });
+
+    add(&Flags::user,
+        "user",
+        "The user this command should run as.");
```
kaysoky: The framework is allowed to specify the user.  However, we have ACLs to restrict users:
```
  message RunTask {
    // Subjects: Framework principals.
    required Entity principals = 1;

    // Objects: Users to run the tasks/executors as.
    required Entity users = 2;
  }
```

kaysoky:
@adam-mesos Renaming complete!

journald/lib_journald.hpp
adam-mesos: Well, but don't the master/agent logs also go to journald? So "mesos journald logger" is still ambiguous.

journald/lib_journald.hpp
adam-mesos: Fair enough. We're hoping that journald will be the default in the future, and dcos is temporarily overriding that.

journald/lib_journald.hpp
adam-mesos: I'm still confused. Is it the case that `Flags` is loaded at module creation and acts as global defaults, whereas `LoggerFlags` are loaded prior to executor launch and can override the global defaults?
If so, it'd be nice if the comment mentioned the class names instead of generically saying "flags" several times.

journald/lib_journald.hpp
adam-mesos: SGTM, glad we don't allow overrides for companion_dir or worker_threads.

journald/journald.cpp
adam-mesos: Alright. Thanks for the explanation. We can merge it with the duplication, but you're going to have to keep an eye on them to make sure they stay in sync.

journald/journald.cpp
adam-mesos: ok, sure

journald/journald.cpp
adam-mesos: I agree that we don't want to add any more options; I'm just sharing a general recommended practice, since you never know what somebody's going to hack onto your module years after you're gone. I feel like it's safer (more future-proof) to match options in the inclusion/exclusion logic to the options you're actually trying to include/exclude, rather than inverting the logic just to save a comparison. I'll let this slide though if you're going to fight it.

journald/journald.cpp
adam-mesos: In that case, we should probably s/journald/logrotate/ in that error message.

journald/journald.cpp
```diff
@@ -129,7 +133,8 @@ class JournaldLoggerProcess : public Process<JournaldLoggerProcess>
           return Nothing();
         }
 
-        if (flags.destination_type != "sandbox") {
+        if (flags.destination_type == "sandbox" ||
+            flags.destination_type == "both") {
```
adam-mesos: Shouldn't this be `if(journald || both)`, since it used to be `!sandbox`?

journald/journald.cpp
```diff
@@ -138,7 +143,8 @@ class JournaldLoggerProcess : public Process<JournaldLoggerProcess>
           }
         }
 
-        if (flags.destination_type != "journald") {
+        if (flags.destination_type == "journald" ||
+            flags.destination_type == "both") {
```
adam-mesos: Shouldn't this be `sandbox || both` since it used to be `!journald`?

adam-mesos: s/sandbox/logrotate/?

journald/journald.cpp
adam-mesos: Nevermind, I see you tried to make the swap. Thanks!

kaysoky:
Whoops 😓 

This only passed tests because the tests only covered `journald+logrotate`.  I've parameterized the tests to cover the other two cases.

tests/journald_tests.cpp
```diff
@@ -270,15 +290,30 @@ TEST_F(JournaldLoggerTest, ROOT_LogToJournald)
...
   ASSERT_TRUE(os::exists(sandboxDirectory));
 
-  string stdoutPath = path::join(sandboxDirectory, "stdout");
-  ASSERT_TRUE(os::exists(stdoutPath));
+  std::string stdoutPath = path::join(sandboxDirectory, "stdout");
+
+  if (GetParam() == "journald") {
+    // Check that the sandbox was *not* written to.
+    // TODO(josephw): This file exists as other parts of the agent will
+    // create it.  We should invert this assertion when we fix this in Mesos.
```
adam-mesos: I noticed these empty stdout/stderr files in the sandbox. What creates these? Is there a MESOS JIRA for preventing that?

tests/journald_tests.cpp
```diff
@@ -270,15 +290,30 @@ TEST_F(JournaldLoggerTest, ROOT_LogToJournald)
...
+    ASSERT_TRUE(os::exists(stdoutPath));
 
-  Result<string> stdout = os::read(stdoutPath);
-  ASSERT_SOME(stdout);
-  EXPECT_TRUE(strings::contains(stdout.get(), specialString));
+    Result<std::string> stdout = os::read(stdoutPath);
+    ASSERT_SOME(stdout);
+    EXPECT_FALSE(strings::contains(stdout.get(), specialString));
```
adam-mesos: Is stdout going to be anything other than an empty string? Can we check for that instead?

tests/journald_tests.cpp
kaysoky: The fetcher and docker containerizer create the empty files:
* https://github.com/apache/mesos/blob/1.1.x/src/slave/containerizer/fetcher.cpp#L740-L767
* https://github.com/apache/mesos/blob/1.1.x/src/slave/containerizer/docker.cpp#L251-L263

As for a JIRA... https://issues.apache.org/jira/browse/MESOS-6843

tests/journald_tests.cpp
kaysoky: It should be empty, but in case we have extremely verbose logging enabled (in the command to run the tests), the file may be non-empty, even with nothing to fetch.

PR #10, opened 2017-01-24
1.8.x

asridharan:
Removed the journald module.


asridharan:
@kaysoky can you go through the `Makefile.am` and `configure.ac` changes to see if the PR makes sense. Have DC/OS pull request against this branch where all integration tests seem to be passing.

Makefile.am
```diff
@@ -148,56 +121,41 @@ libmesos_tests_la_LDFLAGS =				\
...
-test_journald_LDADD =					\
+test_dockercfg_LDADD =					\
   $(MESOS_LDFLAGS)					\
   $(MESOS_BUILD_DIR)/$(BUNDLE_SUBDIR)/.libs/libgmock.la	\
   $(MESOS_BUILD_DIR)/src/.libs/libmesos.la		\
   libmesos_tests.la					\
-  libjournaldlogger.la
-
-# Test (make check) binary for the overlay modules.
-check_PROGRAMS += test-overlay
```
kaysoky: The diff is a bit misleading, due to the re-rearrangement of the test binaries.  If you un-rearranged them, I think the diff will be much more concise.

asridharan:
@kaysoky re-arranged the Makefile.am. Should be more readable now.

asridharan:
Removed the systemd references from configure.ac 

@kaysoky ^^

PR #14, opened 2017-05-19
fixes build issue with gcc 4.8+

GoelDeepak:


bbannier:
Since stout is part of Mesos I am not sure we'd want to suppress all warnings from it instead of fixing them. What's the issue this patch addresses?

GoelDeepak:
it was complaining about auto_ptr being deprecated

asridharan:
@bbannier this issue was a fixed a while back in Mesos:
https://issues.apache.org/jira/browse/MESOS-3799

So needs to be fixed in the `mesos-modules` as well since we are directly inheriting the headers.

configure.ac
```diff
@@ -139,7 +139,8 @@ else
     MESOS_CPPFLAGS+=" -I${mesos_root}/3rdparty/libprocess/include"
 
     # Add stout to include path.
-    MESOS_CPPFLAGS="${MESOS_CPPFLAGS} -I${mesos_root}/3rdparty/stout/include"
+    MESOS_CPPFLAGS="${MESOS_CPPFLAGS} -isystem ${mesos_root}/3rdparty/stout/include"
```
asridharan: @GoelDeepak is it possible to make it specific to `boost`. I think that is where the error was originating from. This is what was done for as well:
https://issues.apache.org/jira/browse/MESOS-3799

GoelDeepak:
@asridharan  I earlier did it only for boost but then stout complained about some other stuff forgot what was it about.

kaysoky:
What are the actual error messages you're seeing in stout?

GoelDeepak:
It was again complaining about auto_ptr

tillt:
Seems this PR is in limbo without clear motivation and reviews. It sounds a lot as if the motivation was not explained sufficiently.

Can we have
1. ticket showing the error/s in full
2. agree on the most appropriate solution for (1)
3. have a PR narrow and verbose enough to solve just that

GoelDeepak:
The complete error can be found https://jira.mesosphere.com/browse/DCOS_OSS-1422

urbanserj:
The issue was fixed as a part of https://github.com/dcos/dcos-mesos-modules/pull/36, please see https://github.com/dcos/dcos-mesos-modules/pull/36/files#diff-67e997bcfdac55191033d57a16d1408aR224

PR #16, opened 2017-06-15
Updated Makefile.am to inherit `MESOS_CXXFLAGS` properly.

tillt:
Without this patch, the configure time setup of `enable-debug` as well as `enable-optimize` will not influence the build results.


tillt:
See https://mesosphere.slack.com/archives/C02JTPBF3/p1497486354488227

PR #20, opened 2017-06-29
Ipv6 support

GoelDeepak:
This change is [Reviewable](https://reviewable.io/reviews/dcos/dcos-mesos-modules/20)

### DIFF
--- a/overlay/master.cpp
+++ b/overlay/master.cpp
@@ -286,11 +286,11 @@ struct Overlay
       Network startSubnet = Network(IP(startAddr), prefix.get());
       Network endSubnet = Network(IP(endAddr), prefix.get());
 
-      LOG(INFO) << "IPv4: " << startSubnet << " - " << endSubnet;
+      LOG(INFO) << name << " IPv4: " << startSubnet << " - " << endSubnet;
 
       freeNetworks +=
         (Bound<Network>::closed(startSubnet),
-         Bound<Network>::open(endSubnet));
+         Bound<Network>::closed(endSubnet));
     }
 
     // IPv6
@@ -312,7 +312,7 @@ struct Overlay
 
       freeNetworks6 +=
         (Bound<Network>::closed(startSubnet6),
-         Bound<Network>::open(endSubnet6));
+         Bound<Network>::closed(endSubnet6));
     }
   }
 
@@ -452,7 +452,7 @@ struct Overlay
       Network startSubnet = Network(IP(startAddr), prefix.get());
       Network endSubnet = Network(IP(endAddr), prefix.get());
 
-      LOG(INFO) << "Reset IPv4: " << startSubnet << " - " << endSubnet;
+      LOG(INFO) << name << " Reset IPv4: " << startSubnet << " - " << endSubnet;
 
       freeNetworks +=
         (Bound<Network>::open(startSubnet),
@@ -476,7 +476,8 @@ struct Overlay
       Network startSubnet6 = Network(IP(startAddr6), prefix6.get());
       Network endSubnet6 = Network(IP(endAddr6), prefix6.get());
 
-      LOG(INFO) << "Reset IPv6: " << startSubnet6 << " - " << endSubnet6; 
+      LOG(INFO) << name 
+                << " Reset IPv6: " << startSubnet6 << " - " << endSubnet6; 
 
       freeNetworks6 +=
         (Bound<Network>::open(startSubnet6),
@@ -580,7 +581,7 @@ public:
     // agent. Queue the message on the Agent. Finally, ask Master to
     // reliably send these messages to the Agent.
     foreachpair (const string& name, const Owned<Overlay>& overlay, _overlays) {
-      // skip if the overlay is present
+      // Skip if the overlay is present
       if (overlays.contains(name)) {
         continue;
       }
@@ -718,7 +719,7 @@ private:
     const string name = _overlay->info().name();
 
     // IPv4
-    Option<Network> network_ = None();
+    Option<Network> _network = None();
     if (_overlay->has_subnet()) {
       Try<Network> network = Network::parse(
           _overlay->subnet(),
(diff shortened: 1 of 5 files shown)


overlay/master.cpp
```diff
@@ -257,21 +252,21 @@ struct Vtep
 
   void reset()
   {
-    freeIP = IntervalSet<IPWrapper>();
+    freeIP = IntervalSet<IP>();
```
asridharan: We still going to do the `.begin` and `.end` methods to simply IPv4 vs IPv6 logic here?

overlay/network.hpp
```diff
@@ -102,50 +113,86 @@ class IPWrapper
...
+// Returns the string representation of the given IP using the
+// canonical form, for example: "10.0.0.1" or "fe80::1".
+inline std::ostream& operator<<(std::ostream& stream, const IP& ip)
+{
+  stream << ip;
   return stream;
 }
 
-class Network
+class Network : public net::IPNetwork
```
asridharan: After upgrading to the latest `libmesos` this will have to change to `net::IP::Network`.

overlay/master.cpp
```diff
@@ -761,24 +963,53 @@ class ManagerProcess : public ProtobufProcess<ManagerProcess>
...
-        uint32_t startIP = ntohl(network.address().in().get().s_addr);
-
-        uint32_t mask = ntohl(network.netmask().in().get().s_addr);
-        mask = ~mask;
-
-        uint32_t endIP = startIP | mask;
+      [&addressSpace](const Network &network) -> Try<Nothing> {
+        IP startIP, endIP;
+        switch(network.address().family()) {
+          case AF_INET: {
```
asridharan: Feel like `startIP` and `endIP` should be `start` and `end` methods in `overlay::Network`. Would really cleanup this code here.

overlay/master.cpp
```diff
@@ -812,33 +1043,67 @@ class ManagerProcess : public ProtobufProcess<ManagerProcess>
...
-      if (address.isError()) {
-        return Error(
-            "Unable to determine subnet for network: " +
-            stringify(address.get()));
+      // IPv4
+      Option<uint8_t> prefix = None();
+      Option<Network> address = None();
+      if (overlay.has_subnet()) {
+        Try<Network> _address =
+          Network::parse(overlay.subnet(), AF_INET);
```
asridharan: Need to leave a line after every multi-line statement.

overlay/master.cpp
```diff
@@ -812,33 +1043,67 @@ class ManagerProcess : public ProtobufProcess<ManagerProcess>
...
+        prefix = overlay.prefix();
       }
 
-      Try<Nothing> valid = updateAddressSpace(address.get());
+      // IPv6
+      Option<uint8_t> prefix6 = None();
+      Option<Network> address6 = None();
+      if (overlay.has_subnet6()) {
+        Try<Network> _address6 = 
+          Network::parse(overlay.subnet6(), AF_INET6);
```
asridharan: Need to leave a line after every multi-line statement.

overlay/master.cpp
```diff
@@ -1239,39 +1548,62 @@ class ManagerProcess : public ProtobufProcess<ManagerProcess>
...
+              << network6.error();
+            abort();
+          }
 
-          abort();
+          LOG(INFO) << "reserving IPv6 " << stringify(network6.get());
+          Try<Nothing> result = overlays.at(overlay.info().name())->reserve6(network6.get());
+          if (result.isError()) {
+            LOG(ERROR) << "Unable to reserve the IPv6 subnet " << network6.get()
+                     << ": " << result.error();
```
asridharan: Minor nit: Align `<<` across the two lines.

overlay/master.cpp
asridharan: Looks like we already have implementation of `begin` and `end` in `overlay::Network`. So we should be able to use it over here?

overlay/network.hpp
```diff
@@ -0,0 +1,415 @@
...
+
+  bool operator>(const IP& that) const
+  {
+    return net::IP::operator>(that);
+  } 
+
+  IP& operator++() {
+    switch (family_) {
+      case AF_INET: {
+       uint32_t address = ntohl(storage_.in_.s_addr);
```
asridharan: The alignment seems a bit off here?

overlay/network.hpp
```diff
@@ -0,0 +1,415 @@
...
+    return *this;
+  }
+
+  uint8_t prefix() const { return prefix_; }
+
+private:
+  uint8_t prefix_;
+  uint8_t index_;
+};
+
```
asridharan: I think we need to insert a new line between method implementations.

tests/overlay_tests.cpp
```diff
@@ -101,11 +101,15 @@ namespace tests {
...
 constexpr char AGENT_JSON_CONFIG[] = "agent.json";
 constexpr char OVERLAY_SUBNET[] = "192.168.0.0/16";
+constexpr char OVERLAY_SUBNET6[] = "fd02::/64";
 constexpr char OVERLAY_NAME[] = "mz-overlay";
 constexpr char MASTER_JSON_CONFIG[] = "master.json";
 constexpr char MASTER_OVERLAY_MODULE_NAME[] =
   "com_mesosphere_mesos_OverlayMasterManager";
 
+constexpr uint32_t OVERLAY_PREFIX = 24;
+constexpr uint32_t OVERLAY_PREFIX6 = 96;
```
asridharan: For docker bridges the minimum expected prefix is 80 I think? Can you double check.

overlay/master.cpp
```diff
@@ -202,100 +233,176 @@ struct Vtep
...
 
-    // Retrieve the integer representation of subnet in the
-    // `IntervalSet`.
-    _subnet &= ~netmask;
-    _subnet = _subnet >> (32 - subnet.prefix());
+  Try<Nothing> reserve(const Network& subnet)
+  {
+    uint32_t _mask = ntohl(subnet.netmask().in().get().s_addr);
+    uint32_t _address = ntohl(subnet.address().in().get().s_addr);
+    Network _subnet = Network(IP(_address & _mask), subnet.prefix());
```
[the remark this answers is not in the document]
asridharan: @GoelDeepak the mask is the same as `subnet.prefix`? So a bit confusing why are recreating the mask here?

PR #23, opened 2017-07-15
Added configurable limit to the size of labels passed in logger.

kaysoky:
This adds a rather primitive limit to the number or size of
labels passed from tasks to the container logger companion
binary.  Once the JSON-ified size of labels surpasses a
configurable size, the module simply cuts off any more
arbitrary labels before proceeding with the logger module's
business as usual.

I would prefer to land this after #21 and #22 if possible.

### DIFF
--- a/journald/lib_journald.cpp
+++ b/journald/lib_journald.cpp
@@ -14,6 +14,7 @@
 #include <process/process.hpp>
 #include <process/subprocess.hpp>
 
+#include <stout/bytes.hpp>
 #include <stout/jsonify.hpp>
 #include <stout/try.hpp>
 #include <stout/nothing.hpp>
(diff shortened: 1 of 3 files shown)


journald/lib_journald.cpp
```diff
@@ -55,6 +56,10 @@ bool enabled();
 namespace mesos {
 namespace journald {
 
+// The number of characters in `{"key":"",value":""},` which is the
+// minimum size of any label when serialized into JSON.
+const Bytes LABEL_PADDING_SIZE = Bytes(21);
```
tillt: You missed one quote before `value` and hence it is 22 bytes.

How about this instead:
```
const Bytes LABEL_PADDING_SIZE = Bytes(string("{\"key\":\"\",\"value\":\"\"},").size());
```

PR #25, opened 2017-07-18
Changes CNI configuration file extention from .cni to .conf

GoelDeepak:


overlay/agent.cpp
```diff
@@ -690,7 +690,7 @@ Future<Nothing> ManagerProcess::configureMesosNetwork(const string& name)
   };
 
   Try<Nothing> write = os::write(
-      path::join(cniDir, name + ".cni"),
+      path::join(cniDir, name + ".conf"),
```
asridharan: During an upgrade is it guaranteed that we remove the CNI config directory? If not we will end up having to CNI configs with the same network name which will cause issues?

GoelDeepak: AFAIK /opt/mesosphere directory is nuked during upgrade

PR #26, opened 2017-07-19
Allow building of the repo in parts on macOS.

tillt:
- adds configure time macOS detection
- makes the journald related things linux only


kaysoky:
What's the reason for adding this compatibility?  The modules will never be used on OSX (as far as I know).

tillt:
@kaysoky my motivation was that I wanted them to be buildable on macOS so that developers can at least tinker with them locally on their Apple machines. My hope is that if things are easy to build, new build issues (e.g. due to Mesos being bumped) can get caught and fixed quicker.

PR #27, opened 2017-08-08
Fixed overlay compilation errors and workaround to UPID optimization

asridharan:
* Fixed compilation error during to introduction of `State` variable in `ProcessBase`.
* Fixed linter errors in master.cpp and agent.cpp.
* Fixed linter to not check the license header since the license is already added as part of the GitHub repo. Also, the license checks were Mesos specific so didn't make sense to make it part of our linter.
* Had to add modified files from `libtool`/m4 directory.

---
This change is [Reviewable](https://reviewable.io/reviews/dcos/dcos-mesos-modules/27)

### DIFF
--- a/tests/overlay_tests.cpp
+++ b/tests/overlay_tests.cpp
@@ -1,38 +1,3 @@
-/**
- * This file is © 2016 Mesosphere, Inc. (“Mesosphere”).  Mesosphere licenses
- * this file to you solely pursuant to the following terms (and you may not use
- * this file except in compliance with such terms):
- *
- * 1. Subject to your compliance with the following terms, Mesosphere hereby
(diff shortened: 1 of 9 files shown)


tests/overlay_tests.cpp
```diff
@@ -157,7 +123,10 @@ class OverlayTest : public MesosTest
                          ".libs/libmesos_network_overlay.so"));
             writer->field("modules", [](JSON::ArrayWriter* writer) {
               auto masterModuleConfig = [](JSON::ObjectWriter* writer) {
-              writer->field("name", "com_mesosphere_mesos_OverlayMasterManager");
+              writer->field(
+                "name",
```
jieyu: indent should be 4 here.

PR #28, opened 2017-08-14
Fixed compilation errors due to immutability of UPID in Mesos.

asridharan:


kaysoky:
I'm sort of surprised that there are only changes in the overlay modules :smiley:

LGTM, but I haven't given it a compile yet.

asridharan:
I think the reason is that its only the overlay module that talks to actors on a remote node, and hence has to manipulate the UPID to formulate the PID for the remote actor. For all the the other modules I think the dispatch is only between local actors.

PR #30, opened 2017-08-22
Fixed CHECK failure for one case with nested containers.

kaysoky:
If a nested container is launched on a parent container after
restarting the agent (the parent container is launched prior to
restart), then the ExecutorInfo passed into the ContainerLogger
may be blank.  This means that fields that are normally required
such as FrameworkID and ExecutorID will be blank.

This includes one test which:
1. Launches a parent container (prints then sleeps)
2. Emulates an agent failover
3. Launches a nested container (prints then exits)
4. Checks that filtering by `AGENT_ID` shows parent+child messages
5. Checks that filtering by `FRAMEWORK_ID` only shows the parent message
```
sudo ./test-journald --gtest_filter="*ROOT_CGROUPS_LaunchThenRecoverThenLaunchNested" --gtest_repeat=100 --gtest_break_on_failure
```

### DIFF
--- a/journald/lib_journald.cpp
+++ b/journald/lib_journald.cpp
@@ -163,10 +163,16 @@ public:
 
     // NOTE: This field is required by the master/agent, but the protobuf
     // is optional for backwards compatibility.
-    CHECK(executorInfo.has_framework_id());
-    label.set_key("FRAMEWORK_ID");
-    label.set_value(executorInfo.framework_id().value());
-    labels.add_labels()->CopyFrom(label);
+    //
+    // NOTE: It is possible for the ExecutorInfo object to be blank if
+    // a nested container is launched after restarting the Mesos agent.
+    // This is because the agent does not need to keep the ExecutorInfo
+    // checkpointed after it has already launched the executor.
+    if (executorInfo.has_framework_id()) {
+      label.set_key("FRAMEWORK_ID");
+      label.set_value(executorInfo.framework_id().value());
+      labels.add_labels()->CopyFrom(label);
+    }
 
     label.set_key("EXECUTOR_ID");
     label.set_value(executorInfo.executor_id().value());
--- a/tests/journald_tests.cpp
+++ b/tests/journald_tests.cpp
@@ -1,3 +1,4 @@
+#include <map>
 #include <string>
 #include <vector>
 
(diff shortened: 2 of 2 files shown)


tests/journald_tests.cpp
```diff
@@ -292,6 +293,189 @@ TEST_F(JournaldLoggerTest, ROOT_LogToJournaldWithBigLabel)
...
+      {"journalctl",
+       "FRAMEWORK_ID=" + stringify(executor.framework_id())});
+
+  AWAIT_READY(firstQuery);
+  EXPECT_TRUE(strings::contains(firstQuery.get(), specialParentString));
+  EXPECT_TRUE(strings::contains(firstQuery.get(), specialChildString));
+
+  AWAIT_READY(secondQuery);
+  EXPECT_TRUE(strings::contains(secondQuery.get(), specialParentString));
+  EXPECT_FALSE(strings::contains(secondQuery.get(), specialChildString));
```
adam-mesos: So, the expectation is that this test would have passed even before your CHECK->if change, but that it will start to fail once we've "fixed" the bug in Mesos?

journald/lib_journald.cpp
```diff
@@ -163,10 +163,16 @@ class JournaldContainerLoggerProcess :
...
     // is optional for backwards compatibility.
-    CHECK(executorInfo.has_framework_id());
-    label.set_key("FRAMEWORK_ID");
-    label.set_value(executorInfo.framework_id().value());
-    labels.add_labels()->CopyFrom(label);
+    //
+    // NOTE: It is possible for the ExecutorInfo object to be blank if
+    // a nested container is launched after restarting the Mesos agent.
+    // This is because the agent does not need to keep the ExecutorInfo
+    // checkpointed after it has already launched the executor.
```
adam-mesos: Is there any race here, or will this reliably happen any time we launch a nested container on a container from the agent's previous start? E.g. restart an agent then try to debug into any UCR container?

tests/journald_tests.cpp
```diff
@@ -292,6 +293,189 @@ TEST_F(JournaldLoggerTest, ROOT_LogToJournaldWithBigLabel)
...
 
 
+// This checks a specific case where the arguments passed into the
+// ContainerLogger will differ if the Mesos agent restarts before
+// launching a nested container.
+TEST_F(JournaldLoggerTest, ROOT_CGROUPS_LaunchThenRecoverThenLaunchNested)
+{
+  mesos::internal::slave::Flags flags = CreateSlaveFlags();
+  flags.launcher = "linux";
+  flags.isolation = "cgroups/cpu,filesystem/linux,namespaces/pid";
```
kaysoky: These are the minimal set of isolators/launchers we typically use for containers that support nesting.  There are a few more that could be enabled and we technically don't need the PID namespaces.  But it doesn't particularly hurt either way.

tests/journald_tests.cpp
```diff
@@ -292,6 +293,189 @@ TEST_F(JournaldLoggerTest, ROOT_LogToJournaldWithBigLabel)
...
+
+  // Use the journald container logger.
+  flags.container_logger = JOURNALD_LOGGER_NAME;
```
kaysoky: All journald logger tests (unless modified otherwise) use the combo logrotate+journald setting.  We run into the problem regardless because the logrotate vs journald setting only has an effect in the companion binary, rather than the logger module loaded into the agent.

tests/journald_tests.cpp
kaysoky: The test would actually have CHECK-failed before the change (earlier in the test, when doing the nested container launch).

After we've fixed the agent recovery, we'll need to tweak this condition.

journald/lib_journald.cpp
kaysoky: This is completely synchronous behavior.  So this the test isn't flaky at all :)

PR #37, opened 2018-04-06
Create testcase for Mesos master failover with DC/OS overlay modules

urbanserj:
Currently, if there are multiple configuraton attempt failures for the overlay module running on the agent, when registered with the overlay module in the Master it leads to the overlay modules actor just giving up on the registration.

We need to test that the number of configuration attemtps are reset when master failover happens, i.e., if a master failover happens and number of config attempts is less than the number of attempts supported by the overlay module the number of config attempts should be reset and the overlay module should continue attempting to register.

JIRA: https://jira.mesosphere.com/browse/DCOS_OSS-728

### DIFF
--- a/overlay/agent.cpp
+++ b/overlay/agent.cpp
@@ -541,6 +541,7 @@ Future<http::Response> ManagerProcess::overlay(const http::Request& request)
 {
   AgentInfo agent;
   agent.set_ip(stringify(self().address.ip));
+  agent.set_configuration_attempts(configAttempts);
 
   foreachvalue (const AgentOverlayInfo& overlay, overlays) {
     agent.add_overlays()->CopyFrom(overlay);
--- a/overlay/agent.cpp
+++ b/overlay/agent.cpp
@@ -109,49 +109,9 @@ static string OVERLAY_HELP()
 
 
 Try<Owned<ManagerProcess>> ManagerProcess::create(
-    const AgentConfig& agentConfig)
+    const AgentConfig& agentConfig,
+    Option<Owned<MasterDetector>> _detector)
 {
-  Option<string> master = None();
-  if (master.isNone() && agentConfig.has_master()) {
-    master = agentConfig.master();
-  } else {
-    master = os::getenv(MESOS_MASTER);
-
-    // If the agent is running as part of the master it will need to
-    // get the ZK URL from 'MESOS_ZK'.
-    if (master.isNone()) {
-      master = os::getenv(MESOS_ZK);
-    }
-  }
-
-  // We should have learned about the master either from the JSON
-  // config of the module, or from the 'MESOS_MASTER' environment
-  // variable.
-  if (master.isNone()) {
-    return Error(
-        "Master unspecified, hence cannot create a "
-        "`MasterDetector`. Please specify master either through "
-        "the JSON config, or 'MESOS_MASTER' environment variable");
-  }
-
-  if (strings::startsWith(master.get(), "file://")) {
-    const string path = master.get().substr(7);
-
-    Try<std::string> read = os::read(path);
-
-    if (read.isError()) {
-      return Error("Error reading master configuration file '" +
-                   path + "': " + read.error());
-    }
-
-    master = read.get();
-  }
-
-  Try<MasterDetector*> detector = MasterDetector::create(master.get());
-  if (detector.isError()) {
-    return Error("Unable to create master detector: " + detector.error());
-  }
-
   Try<Nothing> mkdir = os::mkdir(agentConfig.cni_dir());
   if (mkdir.isError()) {
     return Error("Failed to create CNI config directory: " + mkdir.error());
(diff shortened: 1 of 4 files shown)


tests/overlay_tests.cpp
```diff
@@ -1343,6 +1340,75 @@ TEST_F(OverlayTest, ROOT_checkAddVirtualNetworks)
...
+  AWAIT_READY(agentModule.get()->ready());
+
+  // Simulate a new master detected event to the agent module
+  // StandaloneMasterDetector detector(master.get()->pid);
+  detector->appoint(master.get()->pid);
+
+  // Wait for the agent module to (re-)register.
+  agentRegisteredAcknowledgement = FUTURE_PROTOBUF(
+    AgentRegisteredAcknowledgement(), _, _);
+  AWAIT_READY(agentRegisteredAcknowledgement);
```
asridharan: @urbanserj shouldn't we be exposing the config attempts and checking the number of config attempts?

urbanserj: @asridharan Could you please check the updated version?

https://github.com/dcos/dcos-mesos-modules/blob/d3399b25db3dd2961506a483715695b4559f1ba6/tests/overlay_tests.cpp#L1402-L1421

overlay/agent.cpp
```diff
@@ -207,12 +167,70 @@ Try<Owned<ManagerProcess>> ManagerProcess::create(
...
+Try<Owned<MasterDetector>> ManagerProcess::createDetector(
+    const AgentConfig& agentConfig,
+    Option<Owned<MasterDetector>> _detector)
+{
+  if (_detector.isSome()) {
+      return _detector.get();
+  }
+
+  Option<string> master = None();
+  if (master.isNone() && agentConfig.has_master()) {
```
asridharan: `master` is already `None` so `master.isNone()` check becomes redundant?

overlay/agent.cpp
```diff
@@ -207,12 +167,70 @@ Try<Owned<ManagerProcess>> ManagerProcess::create(
...
         networkConfig,
         agentConfig.max_configuration_attempts(),
-        Owned<MasterDetector>(detector.get())));
+        detector.get()));
+}
+
+
+Try<Owned<MasterDetector>> ManagerProcess::createDetector(
+    const AgentConfig& agentConfig,
+    Option<Owned<MasterDetector>> _detector)
```
asridharan: Lets remove the `_detector` argument, it seems redundant, since we return the `Owned` contained the `Option` if the `Option` is set. We can just check for the `_detector` when we invoke `createDetector`. Please see my comments above.

overlay/agent.cpp
```diff
@@ -207,12 +167,70 @@ Try<Owned<ManagerProcess>> ManagerProcess::create(
     }
   }
 
+  Try<Owned<MasterDetector>> detector =
```
asridharan: Can we do this?
```
auto getDetector = [=]()->Try<Owned<MasterDetector>> {
  if (_detector.isSome()) {
        return _detector.get();
  }

  return ManagerProcess::createDetector(agentConfig);
}

Try<Owned<MasterDetector>> detector = getDetector();
if (detector.isError()) {
  return ....
}

File: overlay/agent.hpp [hunk 5]
```diff
@@ -80,6 +82,10 @@ class ManagerProcess : public ProtobufProcess<ManagerProcess>
       const uint32_t _maxConfigAttempts,
       process::Owned<master::detector::MasterDetector> _detector);
 
+  static Try<process::Owned<master::detector::MasterDetector>> createDetector(
+      const internal::AgentConfig& agentConfig,
+      Option<process::Owned<master::detector::MasterDetector>> detector);
```
asridharan: If we follow the comments above we wouldn't need this argument.

overlay/overlay.proto
```diff
@@ -96,6 +96,10 @@ message AgentInfo {
 
   // The overlay networks that exist on this agent.
   repeated AgentOverlayInfo overlays = 2;
+
+  // Number of configuration attempts. For more information
+  // please see max_configuration_attempts in AgentConfig.
+  optional uint32 configuration_attempts = 3;
```
asridharan: Why do we want this default? The default should be set only for `max_configuration_attemtps`. I think this field gives the state of the number of configuration attempts so it should have a value to start with?

tests/overlay_tests.cpp
```diff
@@ -518,6 +517,7 @@ TEST_F(OverlayTest, checkMasterAgentComm)
 
   AgentInfo masterAgentInfo;
   masterAgentInfo.CopyFrom(state->agents(0));
+  info->clear_configuration_attempts();
```
asridharan: I think we need this only because you are setting a default? If we don't have the default in the protobuf we shouldn't need this?

tests/overlay_tests.cpp
```diff
@@ -810,6 +810,7 @@ TEST_F(OverlayTest, ROOT_checkMasterRecovery)
...
```
asridharan: Same as the comment above.

tests/overlay_tests.cpp
urbanserj: In that test case we checks that agent and master modules have the same configuration. Only agent module has `configuration_attempts` value, so we have to clear it to be able to compare agent module response with master module response.

overlay/overlay.proto
urbanserj: I don't get it. `configuration_attempts` doesn't have default value. it doesn't have any value if it's not set.

overlay/agent.cpp
urbanserj: Thank you for pointing it out! I've removed the second arguments from `createDetector` function and re-write this code.

https://github.com/dcos/dcos-mesos-modules/blob/6649d98e5ee58b20e159f81b967e622f6acb6175/overlay/agent.cpp#L170-L175

overlay/overlay.proto
asridharan: My bad !! Was reading this wrong !!

tests/overlay_tests.cpp
asridharan: 👍

PR #39, opened 2018-04-23
Add MTU configuration for vtep1024

urbanserj:
https://jira.mesosphere.com/browse/DCOS_OSS-1529

> DC/OS Config "dcos_overlay_mtu" can not be used for VTEP1024. VTEP1024's MTU is always greater(more than 54 bytes?) than "dcos_overlay_mtu" for vxlan overhead.

### DIFF
--- a/overlay/master.cpp
+++ b/overlay/master.cpp
@@ -123,10 +123,12 @@ struct Vtep
 {
   Vtep(const Network& _network,
        const Option<Network>& _network6,
-       const MAC _oui)
+       const MAC _oui,
+       const Option<size_t>& _mtu)
     : network(_network),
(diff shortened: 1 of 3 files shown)


overlay/master.cpp
```diff
@@ -1366,6 +1377,9 @@ class ManagerProcess : public ProtobufProcess<ManagerProcess>
       if (vtepIP6.isSome()) {
         vxlan.set_vtep_ip6(stringify(vtepIP6.get()));
       }
+      if (vtep.mtu.isSome()) {
```
asridharan: Minor nit, leave a new line before this line.

PR #40, opened 2018-05-17
Updated the journald logger to the latest logger interface.

jieyu:
This patch set also fixed some test issues.


common/shell.hpp
```diff
@@ -44,7 +44,7 @@ inline process::Future<std::string> runCommand(
       process::io::read(s->out().get()),
       process::io::read(s->err().get()))
     .after(timeout,
-                [&](const process::Future<ProcessTuple> &t) ->
+                [=](const process::Future<ProcessTuple> &t) ->
```
tillt: Thanks!

PR #41, opened 2018-05-17
Made sure runCommand do not return until it is killed.

jieyu:
Previously, it is possible that runCommand returns while the process is
still being killed. This causes the tests to be flaky. This patch fixes
that by waiting for the process to be killed before returning.

### DIFF
--- a/common/shell.hpp
+++ b/common/shell.hpp
@@ -30,58 +30,63 @@ inline process::Future<std::string> runCommand(
 
   if (s.isError()) {
     return process::Failure(
-      "Unable to execute '" + command + "': " + s.error());
+        "Unable to execute '" + command + "': " + s.error());
   }
   pid_t pid = s->pid();
 
   typedef std::tuple<
-    process::Future<Option<int>>,
-    process::Future<std::string>,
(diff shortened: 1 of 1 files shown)


common/shell.hpp
```diff
@@ -30,58 +30,63 @@ inline process::Future<std::string> runCommand(
...
-                [=](const process::Future<ProcessTuple> &t) ->
-                    process::Future<ProcessTuple> {
-        // NOTE: Discarding this future has no effect on the subprocess
-        os::killtree(pid, SIGKILL);
-        return std::make_tuple(-1, "", "timeout after " + stringify(timeout));
+    .after(timeout, [=](const process::Future<ProcessTuple> &future) ->
+        process::Future<ProcessTuple> {
+      // NOTE: Discarding this future has no effect on the subprocess
+      os::killtree(pid, SIGKILL);
+      return future.then(
```
tillt: The process has gone and the reaper is done signaled by those inner futures (`ProcessTuple`) getting into a terminal state. In turn `await`s future gets ready, so we hit `then` - that is when we can finally exit.

PR #42, opened 2018-05-17
Adapted to the new logger interface.

jieyu:
Now the `prepare()` method takes `ContainerID` as a parameter. We no
longer need to calculate the container ID in a hacky way.

### DIFF
--- a/journald/lib_journald.cpp
+++ b/journald/lib_journald.cpp
@@ -3,6 +3,7 @@
 #include <vector>
 
 #include <mesos/mesos.hpp>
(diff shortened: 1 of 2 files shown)


journald/lib_journald.cpp
```diff
@@ -3,6 +3,7 @@
 #include <vector>
 
 #include <mesos/mesos.hpp>
+#include <mesos/type_utils.hpp>
```
tillt: Used for streaming `ContainerID` as done by `stringify(proto::Message)`.

PR #43, opened 2018-05-19
Allowed journald logger to take extra labels from env.

jieyu:
This allows the user to use environment variable to add additional
labels to the journald log entries.


journald/lib_journald.cpp
```diff
@@ -225,14 +251,16 @@ class JournaldContainerLoggerProcess :
     // level containers) , and the executor is named, use that name to
```
tillt: Not yours but could you please kill that extra space before the comma?

journald/lib_journald.cpp
```diff
@@ -225,14 +251,16 @@ class JournaldContainerLoggerProcess :
...
-      label.set_key("SYSLOG_IDENTIFIER");
       label.set_value(
           executorInfo->has_name() ?
           executorInfo->name() :
           executorInfo->executor_id().value());
-      labels.add_labels()->CopyFrom(label);
+    } else {
+      label.set_value("mesos-container");
     }
+    labels.add_labels()->CopyFrom(label);
```
tillt: These are additional changes to those documented in the commit description. Let's make that more transparent in e.g. a ticket and/or in the commit description.

journald/lib_journald.hpp
```diff
@@ -80,6 +83,13 @@ struct LoggerFlags : public virtual flags::FlagsBase
         "    size <logrotate_max_stderr_size>\n"
         "  }\n"
         "NOTE: The 'size' option will be overridden by this module.");
+
+    add(&LoggerFlags::extra_labels,
+        "extra_labels",
+        "Extra kay value pairs (in JSON object format) that will be set\n"
+        "for the journald entry. For example:\n"
```
tillt: Do we need to point out that the  values can only be strings, not numbers or other objects?

tests/journald_tests.cpp
```diff
@@ -551,6 +551,10 @@ TEST_P(JournaldLoggerTest, ROOT_LogToJournald)
   variable->set_name("CONTAINER_LOGGER_DESTINATION_TYPE");
   variable->set_value(GetParam());
 
+  variable = task.mutable_command()->mutable_environment()->add_variables();
+  variable->set_name("CONTAINER_LOGGER_EXTRA_LABELS");
+  variable->set_value("{\"EXTRA_LABEL\":\"extra_label\"}");
```
tillt: Let's also add a label with an invalid value type and confirm it was ignored.

journald/lib_journald.cpp
```diff
@@ -166,6 +172,28 @@ class JournaldContainerLoggerProcess :
...
+      }
+
+      label.set_key(key);
+      label.set_value(_value);
+      labels.add_labels()->CopyFrom(label);
+    }
+
     // NOTE: This field is required by the master/agent, but the protobuf
     // is optional for backwards compatibility.
     //
```
kaysoky: The NOTE right below this line could be updated to match some of the new recovery logic we've added since this comment was added:
```
    // NOTE: It is possible for the ExecutorInfo object to be blank in the
    // following cases:
    //   * Prior to Mesos 1.5.0, the ContainerConfig object is not checkpointed
    //     and hence, any containers recovered from earlier versions will be
    //     missing some metadata.
    //     See: https://issues.apache.org/jira/browse/MESOS-6894
    //   * Standalone containers naturally do not have this field set.
```

journald/lib_journald.cpp
```diff
@@ -215,8 +243,6 @@ class JournaldContainerLoggerProcess :
       labels.add_labels()->CopyFrom(label);
     }
```
kaysoky: The comment above the sandbox directory parsing logic can be updated now:
```
   // Derive the AgentID from the sandbox directory, which always occurs
   // after the `slaves` directory.
   // See: src/slave/paths.hpp in the Mesos codebase for more info.
```

The `ContainerID` is no longer parsed.  And the sandbox directory no longer has to deal with MESOS-1833, because that value comes from the Agent rather than the containerizer.

tests/journald_tests.cpp
```diff
@@ -593,6 +597,10 @@ TEST_P(JournaldLoggerTest, ROOT_LogToJournald)
         {"journalctl",
          "EXECUTOR_ID=" + statusRunning->executor_id().value()});
 
+    Future<std::string> extraLabelQuery = runCommand(
+        "journalctl",
+        {"journalctl", "EXTRA_LABEL=extra_label"});
+
```
kaysoky: This check won't be robust to multiple runs of the same test.  In fact, it would presumably never fail after the first run.

You should add an extra qualifier to the query (like `"FRAMEWORK_ID=" + frameworkId.get().value(),`) to differentiate.  The `ROOT_LogToJournaldWithBigLabel` test does this.

journald/lib_journald.hpp
kaysoky: Maybe you could also mention that this field is meant for cases where ExecutorInfo labels are not applicable, like standalone containers.

journald/lib_journald.cpp
jieyu: Done.

journald/lib_journald.cpp
jieyu: Done.

journald/lib_journald.hpp
jieyu: Done.

journald/lib_journald.cpp
jieyu: Done. Thanks!

journald/lib_journald.cpp
jieyu: Done.

tests/journald_tests.cpp
jieyu: Done! Thanks!

PR #46, opened 2018-06-01
Fixes build issue due to changes in mesos master

GoelDeepak:
Recently, there was a change in mesos that changed await to take
vector instead of list

Mesos commit: a52e5288ea94ec7d84f82c6c01c1ce3c7a3902db


bmahler:
Looks good to me!

PR #47, opened 2018-06-28
Removed unneeded picojson definition.

karya0:
Mesos recently made it unnecessary to manually define
PICOJSON_USE_INT64 to use its stout/json.hpp header. For that they
move the definition into the header itself. We currently still manually
define this flag which leads to a (intentionally made fatal) warning
about a macro redefinition.

This patch removes our defintion.

This PR and description is taken from the corresponding EE PR: https://github.com/mesosphere/dcos-ee-mesos-modules/pull/273

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -344,7 +344,7 @@ AC_CHECK_HEADERS([boost/lexical_cast.hpp boost/functional/hash.hpp],
                 [AC_MSG_ERROR([boost is not installed.])])
 
 AC_CHECK_HEADER([picojson.h],
-                [MESOS_CPPFLAGS+=" -DPICOJSON_USE_INT64 -D__STDC_FORMAT_MACROS"],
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -344,7 +344,7 @@ AC_CHECK_HEADERS([boost/lexical_cast.hpp boost/functional/hash.hpp],
                 [AC_MSG_ERROR([boost is not installed.])])
 
 AC_CHECK_HEADER([picojson.h],
```
bbannier: Checking for `picojson.h` directly seems pretty wrong since if we by some coincidence would pick up a `picojson.h` from the system we would actually be required to define `PICOJSON_USE_INT64` ourself. OTOH, this header is never included directly here, but instead only via stout.

I would remove the check for `picojson.h` here and instead rely on the Mesos installation to provide it.

PR #48, opened 2018-07-14
Adds dataDir in the host-local ipam configuration

GoelDeepak:
Currently, host-local ipam cni plugin uses a persistent location
on the disk as its data dir where it keeps the information about
allocated IP addresses. This leads to orphan IP addresses if
the agent reboots. This patch moves the data dir to a tmpfs location
so that all the allocated IP addresses are recycled upon agent reboot.

jira ticket: DCOS_OSS-3750


overlay/agent.cpp
```diff
@@ -179,9 +179,15 @@ Try<Owned<ManagerProcess>> ManagerProcess::create(
     return Error(detector.error());
   }
 
+  Option<string> cni_data_dir = None();
```
jieyu: nits: no need for `= None()`. default value is None()

PR #50, opened 2018-07-19
Fixes an issue with the iptables rules in Overlay

GoelDeepak:
Docker isolates its networks by adding an explicit DROP iptables rules
for any cross communication. However, Overlay (which uses
docker network) allows seamless communication among Docker networks.
So, Overlay by-passes Docker isolation by adding an explicit RETURN
as the first iptables rule in the chain. However, due to a bug,
Overlay missed adding that rule if Docker network is already
present. This patch fixes that issue

jira ticket: DCOS_OSS-3697

### DIFF
--- a/tests/overlay_tests.cpp
+++ b/tests/overlay_tests.cpp
@@ -105,6 +105,7 @@ constexpr char AGENT_JSON_CONFIG[] = "agent.json";
 constexpr char OVERLAY_SUBNET[] = "192.168.0.0/16";
 constexpr char OVERLAY_SUBNET6[] = "fd02::/64";
 constexpr char OVERLAY_NAME[] = "mz-overlay";
+constexpr char OVERLAY_NAME_2[] = "mz-overlay-2";
 constexpr char MASTER_JSON_CONFIG[] = "master.json";
 constexpr char MASTER_OVERLAY_MODULE_NAME[] =
   "com_mesosphere_mesos_OverlayMasterManager";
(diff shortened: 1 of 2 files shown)


tests/overlay_tests.cpp
```diff
@@ -1731,6 +1740,71 @@ TEST_F(OverlayTest, checkIPv6Configuration)
   ASSERT_EQ(agentOverlay->backend().vxlan().vtep_ip6(), "fd03::1/64");
 }
 
+
+//Test bypass of Docker Isolation
```
jieyu: nits: one space after `//`

tests/overlay_tests.cpp
```diff
@@ -1347,6 +1349,13 @@ TEST_F(OverlayTest, ROOT_checkAddVirtualNetworks)
   ASSERT_EQ(agentOverlay->info().subnet(), "11.0.0.0/8");
   ASSERT_EQ(agentOverlay->subnet6(), "fd04::/80");
   ASSERT_EQ(agentOverlay->info().subnet6(), "fd04::/64");
+
+  // cleanup the network
+  Future<string> docker = runCommand("docker",
```
jieyu: This cleanup should ideally in both TearDown and SetUp in case someone ctrl-c the test.

GoelDeepak: The reason I kept it here is that this network is created by this test.

jieyu: Yeah, i know. I think we can make this network name another constant, and always try to cleanup in the TearDown or SetUp (and ignore errors)

PR #55, opened 2018-08-18
Add the metrics module.

greggomann:
This PR adds the new metrics isolator module. The module does the following:
* Hits the metrics service API during container launch to retrieve a metrics host/port
* Injects the retrieved host/port into the container's environment

### DIFF
--- a/metrics/isolator.cpp
+++ b/metrics/isolator.cpp
@@ -40,6 +40,7 @@ using std::map;
 using std::string;
 using std::vector;
 
+using mesos::slave::ContainerClass;
 using mesos::slave::ContainerLaunchInfo;
 using mesos::slave::ContainerState;
 using mesos::slave::Isolator;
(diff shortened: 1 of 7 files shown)


metrics/isolator.cpp
```diff
@@ -0,0 +1,580 @@
...
+
+namespace mesosphere {
+namespace dcos {
+namespace metrics {
+
+namespace internal {
+
+// TODO(greggomann): Make use of `mesos::internal::serialize` when we get
+// access to 'common/http.hpp' from the Mesos codebase.
+string serialize(
```
kaysoky: It doesn't seem like the other content types (non-JSON) are used anywhere.  Is there a plan or need to support other content types?

tests/metrics_tests.cpp
```diff
@@ -0,0 +1,718 @@
...
+namespace mesos {
+namespace metrics {
+namespace tests {
+
+const string METRICS_PROCESS = "metrics-service";
+const string API_PATH = "container";
+const string LEGACY_HOST = "metricshost";
+const Duration REQUEST_TIMEOUT = Seconds(5);
+
+// A map of container IDs and their associated ports for testing.
```
kaysoky: Extraneous floating comment?

tests/metrics_tests.cpp
```diff
@@ -0,0 +1,718 @@
...
+    statePath = path::join(os::getcwd(), "state");
+    os::mkdir(path::join(statePath, "containers"));
+
+    // Construct the metrics module configuration.
+    const string ip = stringify(process::address().ip);
+    const string port = stringify(process::address().port);
+    const string config =
+"{\n"
+"  \"libraries\": [ {\n"
+"    \"file\": \"" + path::join(MODULES_BUILD_DIR,
```
kaysoky: You could format this with the raw string notation:
```
R"~(
{
  ...
}
)~"
```

tests/metrics_tests.cpp
```diff
@@ -0,0 +1,718 @@
...
+    Try<Nothing> result = ModuleManager::load(modules);
+    ASSERT_SOME(result);
+
+    // Create an instance of the module and store it for testing.
+    Try<Isolator*> isolator_ =
+      ModuleManager::create<Isolator>(
+          "com_mesosphere_dcos_MetricsIsolatorModule");
+
+    ASSERT_SOME(isolator_);
+    isolator = isolator_.get();
```
kaysoky: Is this isolator pointer deallocated anywhere?

tests/metrics_tests.cpp
```diff
@@ -0,0 +1,718 @@
...
+  ASSERT_FALSE(os::exists(statePath));
+}
+
+
+// When the isolator's `recover()` method is called, the legacy state directory
+// has been set in the module configuration, but the state directory does not
+// exist, then the module should return a ready future.
+TEST_F(MetricsTest, RecoverSuccessWithoutLegacyDirectory)
+{
+  Clock::pause();
```
kaysoky: Some of these tests pause the clock, but do not resume them afterwards.

It doesn't seem like there is very much code reliant on timeouts in this test suite though (and any tests which do exercise timeouts have the clock paused), so it should be fine 🤔 .

metrics/isolator.cpp
```diff
@@ -0,0 +1,580 @@
...
+    // (states) or the containerizer (orphans).
+    vector<ContainerID> containers;
+
+    foreach(const ContainerState& state, states) {
+      containers.push_back(state.container_id());
+    }
+
+    foreach(const ContainerID& container, orphans) {
+      containers.push_back(container);
+    }
```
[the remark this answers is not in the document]
kaysoky: Note: It probably isn't a good idea to return a failure if the `cleanup` method fails, because that would abort the other isolators' cleanups, including important ones like the cgroups isolator.

metrics/isolator.hpp
```diff
@@ -0,0 +1,180 @@
...
+// Forward declaration.
+class MetricsIsolatorProcess;
+
+namespace isolator {
+
+struct Flags : virtual flags::FlagsBase
+{
+  Flags()
+  {
+    add(&Flags::service_scheme,
```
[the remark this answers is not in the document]
greggomann: I think I'd prefer to leave this in place for now, since I use it in the tests. When I have some time I could update the tests with a mock metrics service that listens on a domain socket. WDYT?

tests/metrics_tests.cpp
greggomann: Nope! I added a `delete` in the test teardown, thanks!

metrics/isolator.cpp
```diff
@@ -0,0 +1,379 @@
...
+        self(),
+        [=](const Future<http::Response>& response) -> Future<http::Response> {
+          if (!response.isReady()) {
+            return Failure("Error posting 'containerStartRequest' for"
+                           " container '" + containerId.value() + "': " +
+                           (response.isFailed() ?
+                               response.failure() : "Future discarded"));
+          }
+
+          return response.get();
```
kaysoky: Looks like `return response;` would give an identical result, except one less copy perhaps.

metrics/isolator.cpp
```diff
@@ -0,0 +1,379 @@
...
+          }
+
+          // TODO(klueska): For now, we require both `statsd_host` and
+          // `statsd_port` to be set. This may not be true in the future.
+          // We need to revisit this if/when this changes.
+          if (!containerStartResponse->has_statsd_host() ||
+              !containerStartResponse->has_statsd_port()) {
+            const string message = containerStartResponse.isError()
+              ? containerStartResponse.error()
+              : string();
```
kaysoky: I think this ternary is unnecessary since you check for `containerStartResponse.isError()` in the previous conditional and return a Failure.

metrics/isolator.cpp
```diff
@@ -0,0 +1,379 @@
...
+    [](const Parameters& parameters) -> Isolator* {
+      // Convert `parameters` into a map.
+      map<string, string> values;
+      foreach (const Parameter& parameter, parameters.parameter()) {
+        values[parameter.key()] = parameter.value();
+      }
+
+      mesosphere::dcos::metrics::isolator::Flags flags;
+
+      // Load and validate flags from environment and map.
```
kaysoky: Looks like you forgot to update this comment when removing the flags-parsing-from-environment-variables.

metrics/isolator.hpp
```diff
@@ -0,0 +1,164 @@
...
+namespace mesosphere {
+namespace dcos {
+namespace metrics {
+
+// TODO(greggomann): Remove this once we have an equivalent helper in stout.
+// See https://reviews.apache.org/r/68543/.
+template<typename T>
+Try<T> parse(const std::string& body)
+{
+  Try<JSON::Object> json = flags::parse<JSON::Object>(body);
```
kaysoky: Same comment I made in the above review: `JSON::parse<JSON::Object>` would be better.

metrics/messages.proto
```diff
@@ -0,0 +1,34 @@
...
+message ContainerStopRequest {
+  // The ContainerID of the container to
+  // stop listening for metrics from.
+  required string container_id = 1;
+}
+
+message ContainerStopResponse {}
+
+// The format of the legacy state stored by the pre DC/OS 1.12 module.
+message LegacyState {
```
kaysoky: You can delete this object now too, right?

PR #56, opened 2018-08-28
adds -w flag in iptables command

GoelDeepak:


tests/overlay_tests.cpp
```diff
@@ -243,7 +243,7 @@ class OverlayTest : public MesosTest
               masterOverlayConfig.network().overlays()) {
         Future<string> iptables = runCommand(
             "iptables",
-            {"iptables",
```
jieyu: please do `"iptables", "-w"`

tests/overlay_tests.cpp
```diff
@@ -584,7 +584,7 @@ TEST_F(OverlayTest, ROOT_checkMesosNetwork)
 
   // Verify that the `IPMASQ` rules have been installed.
   Future<string> iptables = runCommand("iptables",
-      {"iptables",
+      {"iptables -w",
```
jieyu: Ditto.

tests/overlay_tests.cpp
```diff
@@ -678,7 +678,7 @@ TEST_F(OverlayTest, ROOT_checkDockerNetwork)
...
```
jieyu: Ditto.

tests/overlay_tests.cpp
```diff
@@ -1099,7 +1099,7 @@ TEST_F(OverlayTest, ROOT_checkAgentNetworkConfigChange)
...
```
jieyu: Ditto.

tests/overlay_tests.cpp
```diff
@@ -1112,7 +1112,7 @@ TEST_F(OverlayTest, ROOT_checkAgentNetworkConfigChange)
 
   // Delete the IPMASQ rules.
   iptables = runCommand("iptables",
-      {"iptables",
+      {"iptables -w",
```
jieyu: Ditto

tests/overlay_tests.cpp
```diff
@@ -1144,7 +1144,7 @@ TEST_F(OverlayTest, ROOT_checkAgentNetworkConfigChange)
   // With the `mesos_bridge` and the `docker_bridge` disabled there
   // should be no IPMASQ rules installed. Check IPMASQ rules don't exist.
   iptables = runCommand("iptables",
-      {"iptables",
+      {"iptables -w",
```
jieyu: Ditto

tests/overlay_tests.cpp
```diff
@@ -1786,7 +1786,7 @@ TEST_F(OverlayTest, ROOT_checkDockerIsolation)
 
   // Verify the Docker isolation bypass iptable rule
   Future<string> iptables = runCommand("iptables",
-      {"iptables",
+      {"iptables -w",
```
jieyu: Ditto.

tests/overlay_tests.cpp
```diff
@@ -1795,7 +1795,7 @@ TEST_F(OverlayTest, ROOT_checkDockerIsolation)
 
   // Verify that iptables rule is the first one in the chain
   iptables = runCommand("iptables",
-      {"iptables",
+      {"iptables -w",
```
jieyu: Ditto.

tests/overlay_tests.cpp
```diff
@@ -271,7 +271,7 @@ class OverlayTest : public MesosTest
            "rm",
            OVERLAY_NAME, 
            OVERLAY_NAME_2});
-      AWAIT_READY(docker);
+      docker.await();
```
jieyu: Why this change?

GoelDeepak: many tests cases use only one of the overlay network and not both. In those cases, AWAIT_READY fails when it tries to remove the non-existing overlay network

jieyu: ic. Can you add some comments?

PR #57, opened 2018-09-06
Fix issues in the metrics isolator module.

greggomann:
This PR addresses several issues in the metrics isolator module:
1) The module makes DELETE requests to the metrics service when debug containers are cleaned up, even though we do not make POST requests for those containers when they are launched.
2) There are a couple small typos in logging messages.
3) Some unused protobuf messages have not been removed.


greggomann:
@kaysoky could you review this PR when you have a spare cycle?

PR #58, opened 2018-09-08
fixes iptables rules to support coreos 1800.7.0

GoelDeepak:
Coreos 1800.7.0 comes with default docker version of 18.x which has modified its iptables CHAIN names from `DOCKER-ISOLATION` to `DOCKER-ISOLATION-STAGE-1` and `DOCKER-ISOLATION-STAGE-2` which broke mesos overlay module. This patch detects the CHAIN name before applying the iptables rule so that it could work for both old and newer docker versions.

jira: https://jira.mesosphere.com/browse/DCOS_OSS-3707

### DIFF
--- a/overlay/agent.cpp
+++ b/overlay/agent.cpp
@@ -933,14 +933,21 @@ Future<Nothing> ManagerProcess::__configureDockerNetwork(
   // However, Docker disallows this. So we will install a de-funct
   // rule in the DOCKER-ISOLATION chain to bypass any isolation
   // docker might be trying to enforce.
-  const string iptablesCommand = "iptables -w -D DOCKER-ISOLATION -j RETURN; "
-    "iptables -w -I DOCKER-ISOLATION 1 -j RETURN";
+  const string iptablesCommand = 
+      "iptables -w -D DOCKER-ISOLATION -j RETURN; "
+      "iptables -w -I DOCKER-ISOLATION 1 -j RETURN";
 
   Duration timeout = Milliseconds(networkConfig.command_timeout());
   return runScriptCommand(iptablesCommand, timeout)
-    .then([]() -> Future<Nothing> {
-        return Nothing();
-        });
+      .repair(defer(self(), [timeout](const Future<string>&) {
+          const string iptablesCommand =     
+              "iptables -w -D DOCKER-ISOLATION-STAGE-2 -j RETURN; "
+              "iptables -w -I DOCKER-ISOLATION-STAGE-2 1 -j RETURN";
+          return runScriptCommand(iptablesCommand, timeout);
+      }))
+      .then([]() -> Future<Nothing> {
+          return Nothing();
+      });
 }
 
 ManagerProcess::ManagerProcess(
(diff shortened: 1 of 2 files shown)


overlay/agent.cpp
```diff
@@ -933,14 +933,25 @@ Future<Nothing> ManagerProcess::__configureDockerNetwork(
...
+      "iptables -w -I DOCKER-ISOLATION 1 -j RETURN";
 
   Duration timeout = Milliseconds(networkConfig.command_timeout());
-  return runScriptCommand(iptablesCommand, timeout)
-    .then([]() -> Future<Nothing> {
-        return Nothing();
-        });
+  Future<std::string> response = runScriptCommand(iptablesCommand, timeout);
+
+  response.await();
```
jieyu: This is blocking worker thread. We never do `future.await()` except in tests.

Please use `.then` to do that asynchronously. You can ping Core folks from storage team to have some guidance. cc @bbannier @chhsia0

chhsia0: @GoelDeepak like this:
```cpp
return response
  .recover(defer(self(), [timeout](const Future<string>& future) {
    iptablesCommand = ...;
    return runScriptCommand(iptablesCommand, timeout);
  })
  .then([] { return Nothing(); });
```
The `recover` clause is called if `response` is transitioned into `Failed` or `Discarded`, which returns the result of another `runscriptCommand` retry. If you only want to retry when `response` fails, you can use `repair` instead.

So in the above snippet, if `response` is `Finished`, `response.recover(...)` will return the original future (i.e., `response`); if `response` is `Failed` or `Discarded`, `response.recover(...)` will return a new future (the retry one).

chhsia0: Moreover, you can further do:
```cpp
return runScriptCommand(...)
  .recover(...)
  .then(...);
```

tests/overlay_tests.cpp
```diff
@@ -1786,10 +1786,24 @@ TEST_F(OverlayTest, ROOT_checkDockerIsolation)
...
+  // Check the docker chain
   Future<string> iptables = runCommand("iptables",
       {"iptables", "-w",
-       "-C", "DOCKER-ISOLATION",
+       "-L", dockerChain
+      });
+
+  iptables.await();
+
+  if (!iptables.isReady()) {
```
chhsia0: Let's use `if (iptables.isFailed())` here to match the behavior of `repair`.

tests/overlay_tests.cpp
```diff
@@ -203,8 +203,8 @@ class OverlayTest : public MesosTest
         "iptables -w -t nat -D POSTROUTING -s %s "
         "-m set --match-set %s dst "
         "-j MASQUERADE; "
-        "iptables -w -t filter -D DOCKER-ISOLATION "
-        "-j RETURN; "
+        "iptables -w -t filter -D DOCKER-ISOLATION -j RETURN; "
+        "iptables -w -t filter -D DOCKER-ISOLATION-STAGE-2 -j RETURN; "
```
orsenthil: I don't see why we made a choice of `DOCKER-ISOLATION-STAGE-2` vs `DOCKER-ISOLATION-STAGE-1` here (or not both). 

Does that not matter?

PR #61, opened 2018-09-19
Updated the ::pipe() system calls to pipe2 in lib_journald.

Gilbert88:
This is the followup fix for potential FD leak issue after https://issues.apache.org/jira/browse/MESOS-9151 was fixed in DC/OS.


Gilbert88:
We need to land this PR and backport the fix to DC/OS 1.12, 1.11 and 1.10 branches.

PR #78, opened 2018-11-10
Backported segfault fix in tests and potential badness in general.

tillt:


urbanserj:
Here is the fix for dockercfg_tests.cpp:

https://github.com/dcos/dcos-mesos-modules/commit/31c4d688f9a571b103e73d8ea7be4af02ae109ac#diff-8c7a8f72b4e0dbf2fad3bf8d5bbcd43cR171

PR #79, opened 2018-11-11
Backported fixes for tests that were not prepared for TASK_STARTING.

tillt:
Mostly a cherry-pick of 31c4d688f9a571b103e73d8ea7be4af02ae109ac without including changes to common/shell.hpp as those are not test related.

### DIFF
--- a/tests/journald_tests.cpp
+++ b/tests/journald_tests.cpp
@@ -253,14 +253,19 @@ TEST_F(JournaldLoggerTest, ROOT_LogToJournaldWithBigLabel)
   variable->set_name("CONTAINER_LOGGER_DESTINATION_TYPE");
   variable->set_value("journald");
 
+  Future<TaskStatus> statusStarting;
   Future<TaskStatus> statusRunning;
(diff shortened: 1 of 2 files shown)


tests/journald_tests.cpp
```diff
@@ -294,9 +299,10 @@ TEST_F(JournaldLoggerTest, ROOT_LogToJournaldWithBigLabel)
 }
 
 
-// This checks a specific case where the arguments passed into the
-// ContainerLogger will differ if the Mesos agent restarts before
-// launching a nested container.
+// This test verfies that the executor information will be passed to
```
kaysoky: s/verfies/verifies/

tillt: Dang, I copied that from somewhere - need to fix the original then as well..... Thanks!

tillt:
Seems we need yet another fix as per CI results....

```
../tests/journald_tests.cpp:482
Value of: strings::contains(secondQuery.get(), specialChildString)
  Actual: true
Expected: false
```

PR #80, opened 2018-12-13
dcos6 interface shown in the UI even when enable_ipv6 was set to false

urbanserj:
JIRA: https://jira.mesosphere.com/browse/DCOS-40539
DC/OS PR: https://github.com/dcos/dcos/pull/3939

### DIFF
--- a/overlay/agent.cpp
+++ b/overlay/agent.cpp
@@ -746,7 +746,9 @@ Future<Nothing> ManagerProcess::configureMesosNetwork(const string& name)
     });
     writer->field("chain", strings::upper(overlay.mesos_bridge().name())),
     writer->field("delegate",
-      [subnet, overlay, _cniDataDir, _networkConfig](JSON::ObjectWriter* writer) {
+      [subnet, overlay, _cniDataDir, _networkConfig](
+          JSON::ObjectWriter* writer)
+      {
         writer->field("type", "bridge");
         writer->field("bridge", overlay.mesos_bridge().name());
         writer->field("isGateway", true);
@@ -754,7 +756,9 @@ Future<Nothing> ManagerProcess::configureMesosNetwork(const string& name)
         writer->field("hairpinMode", true);
         writer->field("mtu", _networkConfig.overlay_mtu());
 
-        writer->field("ipam", [subnet, _cniDataDir](JSON::ObjectWriter* writer) {
+        writer->field("ipam", [subnet, _cniDataDir](
+            JSON::ObjectWriter* writer)
+        {
           writer->field("type", "host-local");
           if (_cniDataDir.isSome()) {
             writer->field("dataDir", _cniDataDir.get());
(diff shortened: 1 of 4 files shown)


overlay/agent.cpp
```diff
@@ -351,7 +351,7 @@ void ManagerProcess::updateAgentOverlays(
   // duplicate update corresponding to one already in progress. We
   // should therefore not setup a response for acknowledging this
   // registration.
-  if (futures.empty()) {
+  if (message.overlays_size() > 0 && futures.empty()) {
```
GoelDeepak: Didn't understand why we need this additional condition?

overlay/master.cpp
```diff
@@ -270,12 +270,14 @@ struct Overlay
...
-      const Option<uint8_t> _prefix6)
+      const Option<uint8_t> _prefix6,
+      const bool& _enabled)
     : name(_name),
     network(_network),
     network6(_network6),
     prefix(_prefix),
-    prefix6(_prefix6)
+    prefix6(_prefix6),
+    enabled(_enabled)
```
GoelDeepak: How about `enabled_ip6` or `enabled6` instead of just `enabled`?

overlay/master.cpp
```diff
@@ -331,12 +333,17 @@ struct Overlay
...
       overlay.set_prefix6(prefix6.get());
     }
+    overlay.set_enabled(enabled);
 
     return overlay;
   }
 
   Try<Network> allocate()
   {
+    if (!enabled) {
```
GoelDeepak: There were upgrade issues whenever we tried to skip the allocation. Therefore, by design, the overlay is allocated and the rest of the cluster checks the `flag` to act on it. It is a one-time activity so it should be fine.

overlay/master.cpp
```diff
@@ -349,6 +356,10 @@ struct Overlay
 
   Try<Network> allocate6()
   {
+    if (!enabled) {
```
GoelDeepak: Same as above. Let it allocate

overlay/master.cpp
```diff
@@ -588,6 +602,11 @@ class Agent
         continue;
       }
 
+      // Skip if the overlay is disabled
+      if (!overlay->enabled) {
```
GoelDeepak: Same comment as earlier. Let the system do its normal processing and depend on `flag` to take a call.

tests/overlay_tests.cpp
```diff
@@ -1740,6 +1740,129 @@ TEST_F(OverlayTest, checkIPv6Configuration)
...
+  AWAIT_EXPECT_RESPONSE_HEADER_EQ(
+      APPLICATION_JSON,
+      "Content-Type",
+      agentResponse);
+
+  Try<AgentInfo> info = parseAgentOverlay(agentResponse->body);
+  ASSERT_SOME(info);
+
+  // There should be only one enabled overlay network.
+  ASSERT_EQ(1, info->overlays_size());
```
GoelDeepak: You might have to adjust the test based on the above comments.

overlay/agent.cpp
urbanserj: We expect to get an empty futures list here if there is no overlay network configured or all of them are disabled.

overlay/master.cpp
urbanserj: I would prefer to make this fix not too specific to the ipv6 issue and be able to disable any overlays in general, at least on new agents.

overlay/master.cpp
urbanserj: Could you please elaborate on the issue with upgrade? I can add some tests to check these cases.

The current design is not fine, we all know that. Adding another ipv6 specific flag isn't improving the situation.

GoelDeepak: The system would be in two different internal states if you enable and then disable IPv6 as compare to just keep IPv6 disabled. In the first case, you would see IPv6 configuration in the overlay endpoint whereas in later it won't be present.  This will surely add to the confusion and may lead to certain issues

urbanserj:
Hi @GoelDeepak,

As we discussed in person, I would like to make this patch more generic than just the case with `dcos6` overlay network. I've checked manually on miniDC/OS cluster that my patch works well when ipv6 is disabled -> enabled -> disabled -> enabled, enabled -> disabled -> enabled -> disabled. I was adding new agents nodes in all of these cases; dcos-net and mesos overlay module with the patch handle all of the cases well.

Please let me know if you have any questions.

Thank you.

PR #82, opened 2019-01-04
Add replicated log timeout.

urbanserj:
JIRA: https://jira.mesosphere.com/browse/DCOS_OSS-4575

### DIFF
--- a/overlay/master.cpp
+++ b/overlay/master.cpp
@@ -1214,6 +1214,17 @@ public:
           new mesos::state::protobuf::State(storage));
     }
 
+    // Parse replicatedLogTimeout.
+    Try<Duration> replicatedLogTimeout = Minutes(5);
+    if (masterConfig.has_replicated_log_timeout()) {
+      replicatedLogTimeout =
+        Duration::parse(masterConfig.replicated_log_timeout());
+      if (replicatedLogTimeout.isError()) {
+        return Error("Error parsing replicatedLogTimeout: " +
+                     replicatedLogTimeout.error());
+      }
+    }
+
     return Owned<ManagerProcess>(new ManagerProcess(
           overlays,
           vtepSubnet.get(),
@@ -1222,6 +1233,7 @@ public:
           vtepMTU,
           networkConfig,
           replicatedLog,
+          replicatedLogTimeout.get(),
           storage,
           log));
   }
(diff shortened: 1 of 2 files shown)


GoelDeepak:
@urbanserj LGTM. Could you please add a test for it?

urbanserj:
Hey @GoelDeepak,

I have been trying to add a unit test and found out that it will require pretty big code refactoring. Let me explain.

Please see how mesos tests this case [1] using mocked `Storage` [2] that is passed to `registrar` as a parameter [3]. To test the case in overlay module, we will have to make `Storage` or `State` parametrized. To be able to pass mocked class to `ManagerProcess`, we will have to move the class definitions from `master.cpp` to a new separate file as we did it with `agent.cpp` and `agent.hpp`. I started doing that code refactoring, but realized that it will take few days and tons of work which is out of scope for this issue.

[1] https://github.com/apache/mesos/blob/68ee08d/src/tests/registrar_tests.cpp#L1224-L1251
[2] https://github.com/apache/mesos/blob/68ee08d/src/tests/registrar_tests.cpp#L1214-L1221
[3] https://github.com/apache/mesos/blob/68ee08d/src/master/registrar.hpp#L92

GoelDeepak:
> Hey @GoelDeepak,
> 
> I have been trying to add a unit test and found out that it will require pretty big code refactoring. Let me explain.
> 
> Please see how mesos tests this case [1] using mocked `Storage` [2] that is passed to `registrar` as a parameter [3]. To test the case in overlay module, we will have to make `Storage` or `State` parametrized. To be able to pass mocked class to `ManagerProcess`, we will have to move the class definitions from `master.cpp` to a new separate file as we did it with `agent.cpp` and `agent.hpp`. I started doing that code refactoring, but realized that it will take few days and tons of work which is out of scope for this issue.
> 
> [1] https://github.com/apache/mesos/blob/68ee08d/src/tests/registrar_tests.cpp#L1224-L1251
> [2] https://github.com/apache/mesos/blob/68ee08d/src/tests/registrar_tests.cpp#L1214-L1221
> [3] https://github.com/apache/mesos/blob/68ee08d/src/master/registrar.hpp#L92

I see. Let's capture this in a jira ticket to follow up in future.

urbanserj:
@GoelDeepak Here is the ticket https://jira.mesosphere.com/browse/DCOS-47637

PR #85, opened 2019-01-08
Fixed various style issues in 1.11.

tillt:


tests/overlay_tests.cpp
```diff
@@ -1455,8 +1452,9 @@ TEST_F(OverlayTest, ROOT_checkEnableDisableIPv6Configuration)
   json = JSON::parse<JSON::Array>(docker.get());
   ASSERT_SOME(json);
 
-  // Verify that IPv6 is disabled in docker network
-  ipv6Flag = json->values[0].as<JSON::Object>().find<JSON::Boolean>("EnableIPv6");
+  // Verify that IPv6 is disabled in docker network.
+  ipv6Flag =
+      json->values[0].as<JSON::Object>().find<JSON::Boolean>("EnableIPv6");
```
lava: If we apply Mesos style guides for modules, this should be two spaces.

tillt: We do -- not sure why clang-format decided otherwise here :(

PR #88, opened 2019-01-28
Changed the working directory before changing user.

qianzhangxa:
## High-level description

This change fixes the issue that pod cannot be launched with non-root user because `mesos-journald-logger` has no permission to access container's `stdout`/`stderr` in absolute path.

## Corresponding DC/OS or Apache Mesos tickets (obligatory)

These JIRA ticket(s) must be updated (ideally closed) in the moment this PR lands:

  - [DCOS-47501](https://jira.mesosphere.com/browse/DCOS-47501) Pod cannot be launched with non-root user.

### DIFF
--- a/journald/journald.cpp
+++ b/journald/journald.cpp
@@ -11,12 +11,15 @@
 #include <process/io.hpp>
 #include <process/process.hpp>
 
+#include <stout/check.hpp>
 #include <stout/error.hpp>
 #include <stout/exit.hpp>
 #include <stout/nothing.hpp>
+#include <stout/path.hpp>
 #include <stout/strings.hpp>
 #include <stout/try.hpp>
 
+#include <stout/os/chdir.hpp>
 #include <stout/os/pagesize.hpp>
 #include <stout/os/shell.hpp>
 #include <stout/os/su.hpp>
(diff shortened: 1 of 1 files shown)


qianzhangxa:
I tried this fix in a DC/OS enterprise cluster (permissive mode) and verified pod can be launched with non-root user and log rotate works well.

journald/journald.cpp
```diff
@@ -293,6 +298,22 @@ int main(int argc, char** argv)
     LOG(WARNING) << warning.message;
   }
 
+  // Change the current working directory to the container's sandbox directory.
+  // This is to handle the case that the nested container's user is a non-root
+  // user and different from its parent container's user, in which case the
+  // nested container's user (i.e.,`flags.user`) has no permission to access
+  // `flags.logrotate_filename` since it has no permission to traverse its
+  // parent container's sandbox directory whose permissions is 0750.
+  if (flags.logrotate_filename.isSome()) {
```
kaysoky: This is a required option (checked with a flag validation lambda), so I would prefer:
```
CHECK_SOME(flags.logrotate_filename);

Try<Nothing> result = os::chdir(Path(flags.logrotate_filename.get()).dirname());
...
```

qianzhangxa: @kaysoky That's a bit confused to me, if `flags.destination_type` is `journald` rather than `logrotate`, do we still need `flags.logrotate_filename` to be specified? It seems making `flags.logrotate_filename` required in the case of `journald` does not make sense since we do not need to touch that file at all.

kaysoky: For the most part, we can assume that the journald option is never used.  The performance of journald was terrible for task logging and so we default to logrotate instead.

qianzhangxa: Got it, and I have updated the patch accordingly.

I do not have write access, could you please help do the merge?

PR #100, opened 2019-03-03
Removed libtool.m4 as it is generated by autotools.

tillt:
Also commonly ignored via .gitignore.


tillt:
> Do you want to update the gitignore file as well?

No need?!

https://github.com/dcos/dcos-mesos-modules/blob/master/.gitignore#L16

PR #105, opened 2019-03-21
Fluentbit integration

kaysoky:
## High-level description

What features does this change enable? What bugs does this change fix?

This adds an option to pipe log output over TCP, presumably to a fluentbit server running at a specified address.  In order to configure this properly, the fluentbit IP/Port **must** be added to the module parameters.  And the output mode should probably be set to `fluentbit+logrotate`:
https://github.com/dcos/dcos/blob/master/gen/dcos-config.yaml#L390-L412

The unit test does not require fluentbit as the pipeline is really just a TCP dump of log lines (albeit wrapped in JSON objects).

## Changelog automation

[DCOS-50078](https://jira.mesosphere.com/browse/DCOS-50078) Integrate Mesos task logs with fluentbit component.

### DIFF
--- a/journald/journald.cpp
+++ b/journald/journald.cpp
@@ -5,15 +5,19 @@
 
 #include <systemd/sd-journal.h>
 
+#include <process/address.hpp>
 #include <process/dispatch.hpp>
 #include <process/future.hpp>
 #include <process/id.hpp>
 #include <process/io.hpp>
+#include <process/network.hpp>
 #include <process/process.hpp>
 
 #include <stout/check.hpp>
 #include <stout/error.hpp>
 #include <stout/exit.hpp>
+#include <stout/json.hpp>
+#include <stout/jsonify.hpp>
 #include <stout/nothing.hpp>
 #include <stout/path.hpp>
 #include <stout/strings.hpp>
@@ -54,7 +58,8 @@ public:
     const size_t bufferSize = os::pagesize();
 
     if (flags.destination_type == "logrotate" ||
-        flags.destination_type == "journald+logrotate") {
+        flags.destination_type == "journald+logrotate" ||
+        flags.destination_type == "fluentbit+logrotate") {
       // Populate the `logrotate` configuration file.
       // See `Flags::logrotate_options` for the format.
       //
@@ -189,9 +194,20 @@ public:
         }
 
         if (flags.destination_type == "logrotate" ||
-            flags.destination_type == "journald+logrotate") {
+            flags.destination_type == "journald+logrotate" ||
+            flags.destination_type == "fluentbit+logrotate") {
           // Write the bytes to sandbox, with log rotation.
           Try<Nothing> result = write_logrotate(readSize);
+          if (result.isError()) {
+            promise.fail("Failed to write: " + result.error());
+            return Nothing();
+          }
+        }
+
+        if (flags.destination_type == "fluentbit" ||
+            flags.destination_type == "fluentbit+logrotate") {
+          // Write the bytes to sandbox, with log rotation.
+          Try<Nothing> result = write_fluentbit(readSize);
           if (result.isError()) {
             promise.fail("Failed to write: " + result.error());
             return Nothing();
@@ -230,7 +246,6 @@ public:
     // Even if the write fails, we ignore the error.
     return Nothing();
   }
-
 
   // Writes the buffer from stdin to the leading log file.
   // When the number of written bytes exceeds `--logrotate_max_size`,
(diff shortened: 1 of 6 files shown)


branden:
@kaysoky You can use this JIRA: https://jira.mesosphere.com/browse/DCOS-50078

kaysoky:
Rebased and added a unit test to make this mergable.  There should be no breaking changes from the previous WIP.  
@branden @philipnrmn Please feel free to review/rebump your test branches.

journald/journald.hpp
```diff
@@ -170,6 +187,10 @@ struct Flags : public virtual flags::FlagsBase
   Option<std::string> logrotate_filename;
   std::string logrotate_path;
   Option<std::string> user;
+
+  // Only used if the `destination_type` has `fluentbit`.
```
tillt: I would not use backticks for the value (fluentbit) here, instead single or doublequotes.

tests/journald_tests.cpp
```diff
@@ -877,6 +879,184 @@ TEST_P(JournaldLoggerTest, ROOT_LogToJournald)
...
+INSTANTIATE_TEST_CASE_P(
+    LoggingMode,
+    FluentbitLoggerTest,
+    ::testing::Values(
+        std::string("fluentbit"),
+        std::string("fluentbit+logrotate")));
+
+
+// Sets up a TCP to mock the fluentbit component and then loads the
+// ContainerLogger module and runs a task.
```
tillt: Description might additionally mention that the test expects a known log value to reach the Fluent Bit server.

journald/journald.cpp
```diff
@@ -335,6 +412,9 @@ class JournaldLoggerProcess : public Process<JournaldLoggerProcess>
   int num_entries;
   struct iovec* entries;
 
+  // The active connection to the specified fluentbit address.
```
tillt: Why "active"?

journald/journald.hpp
```diff
@@ -156,6 +158,21 @@ struct Flags : public virtual flags::FlagsBase
     add(&Flags::user,
         "user",
         "The user this command should run as.");
+
+    add(&Flags::fluentbit_ip,
+        "fluentbit_ip",
+        "IP of the fluentbit listener for TCP connections.",
```
tillt: s/fluentbit/Fluent Bit/ ?

maybe;
"IP of the Fluent Bit TCP listener." ?

journald/journald.hpp
```diff
@@ -156,6 +158,21 @@ struct Flags : public virtual flags::FlagsBase
...
+          if (value.isNone()) {
+            return Error("--fluentbit_ip is required");
+          }
+
+          return None();
+        });
+
+    add(&Flags::fluentbit_port,
+        "fluentbit_port",
+        "Port of the fluentbit listener for TCP connections.");
```
tillt: "Port of the Fluent Bit TCP listener." ?

journald/journald.hpp
```diff
@@ -156,6 +158,21 @@ struct Flags : public virtual flags::FlagsBase
     add(&Flags::user,
         "user",
         "The user this command should run as.");
+
+    add(&Flags::fluentbit_ip,
+        "fluentbit_ip",
+        "IP of the fluentbit listener for TCP connections.",
+        [](const Option<net::IP>& value) -> Option<Error> {
+          if (value.isNone()) {
+            return Error("--fluentbit_ip is required");
```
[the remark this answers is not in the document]
kaysoky: Yes.  The task can choose to change the destination of its logging.  So even if the module's default is not set to `fluentbit`, we still need to have the information somewhere.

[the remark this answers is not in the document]
tillt: Got it. My approach would have been to leave it optional and error out when the IP was missing while Fluent Bit was selected. But I guess, this will work just fine....

tests/journald_tests.cpp
```diff
@@ -877,6 +879,184 @@ TEST_P(JournaldLoggerTest, ROOT_LogToJournald)
...
+
+  AWAIT_READY(statusFinished);
+  EXPECT_EQ(TASK_FINISHED, statusFinished.get().state());
+
+  driver.stop();
+  driver.join();
+
+  // We expect two connections to be made:
+  // one for stdout and one for stderr.
+  Future<network::inet::Socket> _client1 = fluentbit_server->accept();
```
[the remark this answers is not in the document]
kaysoky: The clients can successfully connect as soon as this line: 
`Try<Nothing> listen = fluentbit_server->listen(2);` is called in the test setup.
`accept()` is just popping connections off a queue.  

But I agree.  It would be more readable to move these up.

journald/journald.cpp
kaysoky: Hmm... I guess the connection could die.  I'll remove the adjective.

tests/journald_tests.cpp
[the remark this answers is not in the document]
kaysoky: Oh, turns out there is a race here, when using the poll socket.  

When you call `accept()` on a poll socket, we poll the socket for a read event.  Calling this twice in a row may cause the polls to conflict with each other, since they are polling the same socket.  So when I moved the `accept()`s up, the second one would fail with:
`(_client2).failure(): Failed to accept: Resource temporarily unavailable`

The reason why the `accept()`s passed when placed later is because two sockets were ready to be accepted.  When placed early, no sockets are ready.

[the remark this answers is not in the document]
kaysoky: A non-racy way to capture these sockets would be:
```
  // We expect two connections to be made with our `fluentbit_server`:
  // one for stdout and one for stderr.
  Future<network::inet::Socket> _client1 = fluentbit_server->accept();
  AWAIT_ASSERT_READY(_client1);
  
  Future<network::inet::Socket> _client2 = fluentbit_server->accept();
  AWAIT_ASSERT_READY(_client2);
```
Placed after the task has entered the STARTING phase.

PR #106, opened 2019-03-29
Switched FileSink to the new LogSink interface (with usecs).

asekretenko:
## High-level description

This PR switches the FileSink in the logsink module to using the new LogSink interface added at https://issues.apache.org/jira/browse/MESOS-9687.

This finally solves the problem with zeroed microseconds in the logs which are written through this module:
https://jira.mesosphere.com/browse/DCOS-24365

Testing done: manually checked that there are microseconds in the DCOS built with this patch (build and DCOS integration tests: https://github.com/dcos/dcos/pull/4934)

## Changelog automation

[DCOS-24365](https://jira.mesosphere.com/browse/DCOS-24365) Switched FileSink to the new LogSink interface (with microseconds).

### DIFF
--- a/logsink/logsink.hpp
+++ b/logsink/logsink.hpp
@@ -47,6 +47,22 @@ public:
       int line,
       const struct ::tm* tm_time,
       const char* message,
+      size_t message_len,
+      google::int32 usecs);
+
+  // This method implements the now obsolete LogSink interface without the
+  // `usecs` argument introduced by a glog patch in MESOS-9687.
+  //
+  // At the moment of writing defining this method is still necessary due
+  // to the need to maintain compatibility between glog and the old LogSink
+  // implementations.
+  virtual void send(
+      google::LogSeverity severity,
+      const char* full_filename,
+      const char* base_filename,
+      int line,
+      const struct ::tm* tm_time,
+      const char* message,
       size_t message_len);
 
   virtual void WaitTillSent();
--- a/logsink/logsink.cpp
+++ b/logsink/logsink.cpp
@@ -60,7 +60,8 @@ void FileSink::send(
     int line,
     const struct ::tm* tm_time,
     const char* message,
-    size_t message_len)
+    size_t message_len,
+    google::int32 usecs)
 {
   os::write(
       logFd,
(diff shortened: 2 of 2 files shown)


logsink/logsink.hpp
```diff
@@ -49,6 +49,16 @@ class FileSink : public google::LogSink
       const char* message,
       size_t message_len);
 
+  virtual void send(
```
greggomann: Could you leave a comment above the old deprecated method declaration explaining the historical reasons for its existence, with mention of the relevant JIRA tickets? (DCOS-24365 and MESOS-9687)

asekretenko: Actually, there are two reasons - added them in comments.

logsink/logsink.hpp
```diff
@@ -40,6 +40,23 @@ class FileSink : public google::LogSink
...
+      int line,
+      const struct ::tm* tm_time,
+      const char* message,
+      size_t message_len,
+      google::int32 usecs);
+
+  // This method implements the now obsolete LogSink interface without the
+  // `usecs` argument introduced by a glog patch in MESOS-9687.
+  //
+  // At the moment of the writing defining this method is still necessary due
```
greggomann: s/the writing/writing/

logsink/logsink.hpp
```diff
@@ -40,6 +40,23 @@ class FileSink : public google::LogSink
...
+  // to the need to maintain compatibility between glog and the old LogSink
+  // implementations. (In the base class this method is purely virtual,
+  // and the new sink() is implemented there in the terms of the obsolete one.)
```
greggomann: I think it's accurate to say the reverse: "the obsolete sink() is implemented there in terms of the new one.", since the older obsolete method now simply calls into the new one.

logsink/logsink.hpp
asekretenko: Thanks! Articles in  germanic languages have always been confusing me, and likely will continue  for a long time ;)

logsink/logsink.hpp
asekretenko: Right now the situations in the base and in the new derived class are exactly opposite to each other: `LogSink::send(..., usecs)` is implemented in terms of `LogSink::send() = 0`, but `FileSink::send()` is defined in terms `FileSink::send(..., usecs)`

But I hope this will change at some point.

I decided to remove this sentence altogether to avoid confusing readers.

PR #112, opened 2019-06-25
Add overlay metrics.

ichernetsky:
## High-level description

Add overlay metrics to ease detection of common issues users of the overlay module encounter.

## Changelog automation

[DCOS_OSS-5321](https://jira.mesosphere.com/browse/DCOS_OSS-5321) Add overlay metrics.

### DIFF
--- a/overlay/master.cpp
+++ b/overlay/master.cpp
@@ -1,6 +1,7 @@
 #include <stdio.h>
 
 #include <list>
+#include <memory>
 
 #include <stout/check.hpp>
 #include <stout/interval.hpp>
@@ -30,6 +31,7 @@
 #include <mesos/state/storage.hpp>
 #include <mesos/zookeeper/detector.hpp>
 
+#include "master_metrics.hpp"
 #include "messages.hpp"
 #include "network.hpp"
 #include "overlay.hpp"
@@ -531,7 +533,8 @@ struct Overlay
 class Agent
 {
 public:
-  static Try<Agent> create(const AgentInfo& agentInfo)
+  static Try<Agent> create(std::shared_ptr<Metrics> metrics,
+                           const AgentInfo& agentInfo)
   {
     Try<IP> _ip = IP::parse(agentInfo.ip(), AF_INET);
 
@@ -545,7 +548,7 @@ public:
       backend = agentInfo.overlays(0).backend();
     }
 
-    Agent agent(_ip.get(), backend);
+    Agent agent(metrics, _ip.get(), backend);
 
     for (int i = 0; i < agentInfo.overlays_size(); i++) {
       agent.addOverlay(agentInfo.overlays(i));
@@ -559,8 +562,10 @@ public:
     return agent;
   }
 
-  Agent(const IP& _ip, const Option<BackendInfo>& _backend = None())
-    : backend(_backend),
+  Agent(std::shared_ptr<Metrics> metrics, const IP& _ip,
+        const Option<BackendInfo>& _backend = None())
+    : metrics(metrics),
+      backend(_backend),
       ip(_ip) {};
   const IP getIP() const { return ip; };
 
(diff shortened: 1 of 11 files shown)


overlay/master_metrics.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+  // Counts the number of IPv4 subnet allocation failures.
+  process::metrics::Counter subnet_allocation_failures;
+
+  // Counts the number of IPv6 subnet allocation failures.
+  process::metrics::Counter subnet6_allocation_failures;
+
+  // Counts the number of network bridge allocation failures.
+  process::metrics::Counter bridge_allocation_failures;
+};
```
urbanserj: Could you please add counter metrics for how many `RegisterAgentMessage`s and `AgentRegisteredMessage`s master module receives, and how many `UpdateAgentOverlaysMessage`s and `AgentRegisteredAcknowledgement`s it sends?

overlay/master.cpp
```diff
@@ -1551,6 +1563,7 @@ class ManagerProcess : public ProtobufProcess<ManagerProcess>
     CHECK_NOTNULL(replicatedLog.get());
 
     recovering = true;
+    metrics->recovering = 1;
```
urbanserj: Master module has three states: stand by, recovering, and recovered. With `recovering` metric, how can we differentiate the first and the last states?

When module is restarted by supervisor module, should we reset `recovering` metric explicitly?

overlay/master_metrics.hpp
ichernetsky: Absolutely.

overlay/master.cpp
ichernetsky: > Master module has three states: stand by, recovering, and recovered. With recovering metric, how can we differentiate the first and the last states?

Hmm, probably I am missing something. My idea is to alert if any replica stays in `recovering` state for too long. By `stand by` mode, do you mean that there is at most active leader, and all other replicas are standing by? If so, this is the part I do not get. I can map the states you listed to certain implementation details of replicated log, which are not exposed to the outside world. Instead, there is an API of different kind, which is used by the overlay master processes. The best I can think of is, introducing `active` metric, which is set to 0 in every master processes upon start-up, which stays that way if there are no agent registration messages received. Once such a message gets received, the recovery process begins, and `recovering` changes accordingly. Once recovery is completed, `active` is set to 1. Are you thinking of something along those lines? If no, could you please elaborate?

> When module is restarted by supervisor module, should we reset recovering metric explicitly?

I don't think so, because the metric gets recreated upon restart anyways.

urbanserj: It looks like we call `recover()` function when we receive the first `RegisterAgentMessage` which is sent by agent module to that master node when it detects that the master node is elected as a new leader. When new election happens, we do noting in master module. It means that we only can set `active` to 1, but never can set it back to `0`.

I think we are fine if you add a counter for `RegisterAgentMessage`s: if the metric is 0 and `recovering` is 0 then we never tried to recover module's state from the replicated log; if the metric is 1 or bigger and `recovering` is 1 then the module is recovered.

ichernetsky:
Team, could you please let me know if these metrics names are fine with you. What do you think of having `overlay/master` and `overlay/slave` prefixes? If you think this distinction makes sense, what do you think about renaming `overlay/log/ensemble_size` and `overlay/log/recovered` to `overlay/master/log/ensemble_size` and `overlay/master/log/recovered` respectively? The Telegraf Mesos plugin doesn't seem to [collect](https://github.com/dcos/telegraf/blob/1.9.4-dcos/plugins/inputs/mesos/mesos.go) them as of now.

urbanserj:
@ichernetsky I really like your idea! btw how do we convert mesos metric names, i.e. what names will be in prometheus?

ichernetsky:
@urbanserj, metrics names are used as is in the plugin itself, and AFAIU, the Telegraf's Prometheus output plugin replaces invalid (from the POV of Prometheus) characters with underscores. Please refer to https://github.com/dcos/telegraf/blob/1.9.4-dcos/plugins/inputs/mesos/mesos.go for further details.

ichernetsky:
The Telegraf plugin adjustments are here: https://github.com/dcos/telegraf/pull/59
This one should (obviously) be merged before the PR against Telegraf repo.

ichernetsky:
@GoelDeepak, do I have your blessing to merge it?

PR #113, opened 2019-09-05
Drop overlay master state records for agents that are gone.

ichernetsky:
## High-level description

What features does this change enable? What bugs does this change fix?

## Changelog automation

[DCOS_OSS-<number>](https://jira.mesosphere.com/browse/DCOS_OSS-<number>) Descriptive CHANGELOG entry title.


ichernetsky:
A GC script is coming next.

README.md
```diff
@@ -31,6 +31,9 @@ cd build
 make
 ```
 
+In order to compile the Mesos modules tests, `--enable-tests-install`
+should be added to the `configure`'s command-line arguments too.
```
tillt: Thanks for this!

tests/overlay_tests.cpp
```diff
@@ -1981,7 +1981,129 @@ TEST_F(OverlayTest, supervisor)
...
+
+  AWAIT_EXPECT_RESPONSE_STATUS_EQ(OK().status, stateResponse);
+  AWAIT_EXPECT_RESPONSE_HEADER_EQ(
+      APPLICATION_JSON,
+      "Content-Type",
+      stateResponse);
+
+  state = parseMasterState(stateResponse->body);
+  ASSERT_SOME(state);
+  ASSERT_EQ(0, state->agents_size());
```
urbanserj: Could you please also add a test case that checks if it's possible to allocate IP that was previously deallocated?

ichernetsky: Done. Please check it out.

urbanserj: I didn't find that it checks if the same subnet is allocated. Did I overlook it?

ichernetsky: You mean, that it gets exactly the same subnet that it got the first time it was tilted up? There is no such guarantee, and AFAIU, there can't be any since it depends on the order in which agent registration requests come in.

urbanserj: For testing purpose we can have a very small subnet for all agents that allows to have only one agent node to be allocated. I'd like to have some test coverage to the whole use case, it's not just deallocation, but also re-use of deallocated subnets by agent nodes with new node ip addresses.

PR #118, opened 2019-10-07
Technology preview: statsd metrics isolator for Windows

tillt:
## High-level description

Ports the metrics isolator to a Windows compatible build.

This is a stop-gap solution until this module will officially be supported on Windows. This is not officially supported by D2iQ but a technology preview.

Removes the support for linux domain sockets on Windows for obvious reasons. It is yet to be determined if our pipeline supports or even mandates named pipes or alike instead.

## Changelog automation

[DCOS-58008](https://jira.mesosphere.com/browse/DCOS-58008) Port the open source metrics module to Windows.

### DIFF
--- a/tests/metrics_tests.cpp
+++ b/tests/metrics_tests.cpp
@@ -99,7 +99,15 @@ protected:
           } ]
         }
         )~",
+
+        // TODO(tillt): We need to rework this for Windows as soon as
+        // there are standalone modules on that platform.
+#ifndef __WINDOWS__
         path::join(MODULES_BUILD_DIR, ".libs", "libmetrics-module.so"),
+#else
+        "C:/fake/location/metrics-module.dll",
+#endif // __WINDOWS__
+
         ip + ":" + port,
         "/" + METRICS_PROCESS + "/" + API_PATH,
         stringify(REQUEST_TIMEOUT));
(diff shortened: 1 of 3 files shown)


tests/metrics_tests.cpp
```diff
@@ -99,7 +99,15 @@ class MetricsTest : public MesosTest
...
         )~",
+
+        // TODO(tillt): We need to rework this for Windows as soon as
+        // there are standalone modules on that platform.
+#ifndef __WINDOWS__
         path::join(MODULES_BUILD_DIR, ".libs", "libmetrics-module.so"),
+#else
+        "C:/fake/location/metrics-module.dll",
+#endif // __WINDOWS__
+
```
kaysoky: I was wondering why this line would not work for this test, whereas other usages of `MODULES_BUILD_DIR` appear fine.  The other tests tend to use `MODULES_BUILD_DIR` to read a templated `modules.json` file.  `os::read` will handle with mixed separators just fine (`MODULES_BUILD_DIR` is generated by CMake with `/`).

We don't have other tests inserting the module library into the module protobuf.

kaysoky: Presumably, there is a JSON parsing error because inserting `MODULES_BUILD_DIR\.libs\libmetrics-module.so` into the JSON includes un-escaped backslashes.

PR #123, opened 2019-10-27
Reserve all agent VTEP IPs upon recovering from replicated log.

ichernetsky:
## High-level description

It is a fix of bug that could lead to double-allocation of the same VTEP IP address to different agents.

## Changelog automation

[DCOS_OSS-5626](https://jira.mesosphere.com/browse/DCOS_OSS-5626) Mesos overlay master must reserve all VTEP IP addresses of all overlays of all agents upon recovering from the replicated log.

### DIFF
--- a/overlay/master.cpp
+++ b/overlay/master.cpp
@@ -1902,6 +1902,8 @@ protected:
       agents.emplace(agent->getIP(), agent.get());
       VLOG(1) << "Recovered agent: " << agent->getIP();
 
+      hashset <string> reservedVTEPIPs, reservedVTEPIP6s;
+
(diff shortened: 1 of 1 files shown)


urbanserj:
@ichernetsky would it make sense to add one more test to cover this case?

ichernetsky:
@urbanserj, it is impossible to write it in a such a way that it is deterministic, because the order in which overlays are stored in protobufs depends on the order of traversal of hashmaps. So there are two options:

- just write a test and then run it, and hope that traversal of hashmaps doesn't simply hide the initial issue.
- replace hashmaps with some insertion order preserving data structure/augument hashmaps with other data structures to make sure that the order is preserved, and then write a test.

PR #128, opened 2019-11-18
Updated tests to follow IFDEF inversion in journald logger.

tillt:
Enables tests that don't need Linux on Windows and macOS.


tillt:
Yoloing - CI liked it - should be fine.

PR #131, opened 2020-02-19
Ported to cmake build system.

akornatskyy:
# Work In Progress

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,204 @@
+
+if (NOT WIN32)
+  cmake_minimum_required(VERSION 3.7.0)
+else ()
+  cmake_minimum_required(VERSION 3.8.0)
+endif ()
+
+project(dcos-mesos-modules)
+
+set(MESOS_ROOT "../mesos" CACHE FILEPATH "Mesos root directory.")
+get_filename_component(MESOS_ROOT ${MESOS_ROOT} ABSOLUTE)
+
+message("MESOS_ROOT is ${MESOS_ROOT}")
+
+# CMAKE MODULE SETUP.
+#####################
+# Paths that are searched when `include(...)` is called.
+list(APPEND CMAKE_MODULE_PATH ${MESOS_ROOT}/cmake)
+list(APPEND CMAKE_MODULE_PATH ${MESOS_ROOT}/3rdparty/cmake)
+list(APPEND CMAKE_MODULE_PATH ${MESOS_ROOT}/src/cmake)
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
+
+# Macros.
+include(CTest)
+include(Versions)
+include(Dependencies)
+
+# Set the default standard to C++11 for all targets.
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+# Do not use, for example, `-std=gnu++11`.
+set(CMAKE_CXX_EXTENSIONS OFF)
+
+if (CMAKE_CXX_COMPILER_ID MATCHES MSVC)
+  # There is corresponding TODO in mesos as well.
+  #
+  # TODO(andschwa): Switch to `/W4` and re-enable possible-loss-of-data warnings.
+  #
+  # The last two warnings are disabled (well, put into `/W4`) because
+  # there is no easy equivalent to enable them for GCC/Clang without
+  # also fixing all the warnings from `-Wconversion`.
+  add_compile_options(
+    # Like `-Wall`; `/W4` is more like `-Wall -Wextra`.
+    /W3
+    # Disable permissiveness.
+    /permissive-
+    # C4244 is a possible loss of data warning for integer conversions.
+    /w44244
+    # C4267 is a possible loss of data warning when converting from `size_t`.
+    /w44267)
+endif ()
+
+if (WIN32)
+  # COFF/PE and friends are somewhat limited in the number of sections they
+  # allow for an object file. We use this to avoid those problems.
+  add_compile_options(/bigobj /vd2)
+
+  # Fix Warning C4530: C++ exception handler used, but unwind semantics are not
+  # enabled.
+  add_compile_options(/EHsc)
+
+  set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
+
+  # Enable multi-threaded compilation for `cl.exe`.
+  add_compile_options(/MP)
+
+  # Force use of Unicode C and C++ Windows APIs.
+  add_definitions(-DUNICODE -D_UNICODE -DNOGDI -DNOMINMAX -DGOOGLE_GLOG_DLL_DECL=)
+
+  # Convenience flags to simplify Windows support in C++ source; used to
+  # `#ifdef` out some platform-specific parts of Mesos.  We choose to define
+  # a new flag rather than using an existing flag (`_WIN32`) because we want
+  # to give the build system fine-grained control over what code is #ifdef'd
+  # out in the future.  Using only flags defined by our build system to control
+  # this logic is the clearest and most stable way of accomplishing this.
+  add_definitions(-D__WINDOWS__)
+endif ()
+
+add_subdirectory(dockercfg)
+add_subdirectory(journald)
+add_subdirectory(logsink)
+
+# TODO(akornatskyy): To make it possible to use cmake build in DCOS on Linux,
+# overlay module must be re-enabled. For this, import/include layout in the
+# module needs to be fixed there while not breaking automake build
+# (see FIXME's in overaly/CMakeLists.txt).
+#add_subdirectory(overlay)
+
+
+# FIXME(akornatskyy): due to limitation of protobuf generation process
+# and include in source
+include_directories(
+  .
+  build
+  "${BOOST_INCL}"
+  "${CURL_INCL}"
+  "${GLOG_INCL}"
+  "${MESOS_ROOT}/build/include"
+  "${MESOS_ROOT}/include"
+  "${PICOJSON_INCL}"
+  "${PROCESS_INCL}"
+  "${PROTOBUF_INCL}"
+  "${RAPIDJSON_INCL}"
+  "${STOUT_INCL}"
+)
+
+add_library(metrics SHARED metrics/isolator.cpp)
+
+set(CMAKE_CURRENT_BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR}/metrics)
+
+protobuf_generate(
+  LANGUAGE cpp
+  TARGET metrics
+  PROTOS metrics/messages.proto
+  APPEND_PATH ./metrics
+)
+
+target_link_libraries(metrics PRIVATE
+  ${OPENSSL_LIBRARIES}
+  glog
+  http_parser
+  mesos
+  mesos_protobufs
+  process
+  protobuf
+  zlib
+  $<$<PLATFORM_ID:Windows>:Mswsock>
+  $<$<PLATFORM_ID:Windows>:Userenv>
+  $<$<PLATFORM_ID:Windows>:ws2_32>
+)
+
+if (BUILD_TESTING)
+  add_subdirectory(tests)
+
+  # Test binary for metrics
+  add_executable(test-metrics tests/metrics_tests.cpp)
+
+  # TODO(akornatskyy): Paths defined as "nowhere" are needed to build Mesos
+  # test utils (https://github.com/apache/mesos/blob/master/src/tests/utils.cpp).
+  # These definitions can be dropped if the test utilities we need here are
+  # split from the path-dependent utilities into a separate file.
+  add_definitions(
+    -DMODULES_BUILD_DIR="${PROJECT_BINARY_DIR}"
+    -DSOURCE_DIR="${MESOS_ROOT}"
+    -DBUILD_DIR="${MESOS_ROOT}/build"
+    -DPKGMODULEDIR="nowhere"
+    -DLIBDIR="nowhere"
+    -DPKGLIBEXECDIR="nowhere"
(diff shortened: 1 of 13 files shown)


CMakeLists.txt
```diff
@@ -0,0 +1,87 @@
...
+# Paths that are searched when `include(...)` is called.
+list(APPEND CMAKE_MODULE_PATH ${MESOS_ROOT}/cmake)
+list(APPEND CMAKE_MODULE_PATH ${MESOS_ROOT}/3rdparty/cmake)
+list(APPEND CMAKE_MODULE_PATH ${MESOS_ROOT}/src/cmake)
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
+
+# Macros.
+include(Versions)
+include(Dependencies)
+#include(cotire)
```
greggomann: Can we get rid of this?

CMakeLists.txt
```diff
@@ -0,0 +1,87 @@
...
+
+add_subdirectory(dockercfg)
+add_subdirectory(journald)
+add_subdirectory(logsink)
+add_subdirectory(metrics)
+
+# FIXME: due to import path in 'overlay/messages.proto'
+# NOTE: according to [1], overlay module is not required for Windows
+# [1] https://github.com/mesosphere/dcos-enterprise/tree/master/packages/mesos/extra/windows-ee.patches
+#add_subdirectory(overlay)
```
greggomann: I think we can get rid of this?

cmake/Dependencies.cmake
```diff
@@ -0,0 +1,82 @@
...
+
+find_library(
+  MESOS_LIB
+  NAMES mesos
+  PATHS ${MESOS_ROOT}/build/src
+  PATH_SUFFIXES Debug Release
+)
+
+SET(Protobuf_DIR ${MESOS_ROOT}/build/3rdparty/protobuf-${PROTOBUF_VERSION}/src/protobuf-${PROTOBUF_VERSION}-build/cmake)
+find_package(Protobuf CONFIG REQUIRED)
```
greggomann: When I attempt to build locally cmake has some issue finding valid config for protobuf, does this work OK for you guys?

overlay/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
+
```
greggomann: If we're not going to include the overlay module in the cmake build for now, perhaps we should just remove this file?

CMakeLists.txt
akornatskyy: preserved for linux

CMakeLists.txt
```diff
@@ -0,0 +1,198 @@
+
+if (NOT WIN32)
+  cmake_minimum_required(VERSION 3.7.0)
+else ()
+  cmake_minimum_required(VERSION 3.8.0)
```
janisz: Can we just stick with 3.8?

akornatskyy: we want to be consistent with mesos build. it uses 3.7.0 for linux.

CMakeLists.txt
```diff
@@ -0,0 +1,198 @@
...
+add_subdirectory(journald)
+add_subdirectory(logsink)
+
+# FIXME: due to import path in 'overlay/messages.proto'
+# NOTE: according to [1], overlay module is not required for Windows
+# [1] https://github.com/mesosphere/dcos-enterprise/tree/master/packages/mesos/extra/windows-ee.patches
+#add_subdirectory(overlay)
+
+
+# FIXME: due to limitation of protobuf generation process
```
asekretenko: Could you please check that all FIXMEs (in both opensource and EE PRs) are either resolved now or converted into  TODOs that clearly state what exactly should be fixed and what the desired state is?

CMakeLists.txt
```diff
@@ -0,0 +1,198 @@
...
+  add_definitions(
+    -DMODULES_BUILD_DIR="${PROJECT_BINARY_DIR}"
+    -DSOURCE_DIR="${MESOS_ROOT}"
+    -DBUILD_DIR="${MESOS_ROOT}/build"
+    -DPKGMODULEDIR="nowhere"
+    -DLIBDIR="nowhere"
+    -DPKGLIBEXECDIR="nowhere"
+    -DTESTLIBEXECDIR="nowhere"
+    -DSBINDIR="nowhere"
+    -DPKGDATADIR="nowhere"
```
asekretenko: Why do we need to define these at all? 

Is it because something (what exactly?) depends on these definitions existing but doesn't actually use a value?
Or is it just because something will break in a non-transparent way?

In any case, I would ask you to leave a comment explaining the need for these definitions, so that readers don't get lost in the future.

metrics/isolator.hpp
```diff
@@ -170,6 +170,7 @@ class MetricsIsolator : public mesos::slave::Isolator
 } // namespace dcos {
 } // namespace mesosphere {
 
-extern mesos::modules::Module<mesos::slave::Isolator> com_mesosphere_dcos_MetricsIsolatorModule;
+extern "C" mesos::modules::Module<mesos::slave::Isolator>
```
asekretenko: Do you have a good explanation why we need "C" linkage here? Is it something related to MSVC, to Windows... or to cmake? 

Can you add a comment or, in case you don't have a good explanation, a TODO so that we figure this out later?

overlay/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+  "${CURL_INCL}"
+  "${ZOOKEEPER_INCL}"
+  "${MESOS_ROOT}/build/include"
+  "${MESOS_ROOT}/include"
+  # FIXME: due to limitation of protobuf generation process
+  # and include in source
+  ${CMAKE_CURRENT_BINARY_DIR}/..
+  # FIXME: due to the following in agent.hpp
+  # #include <overlay/agent_metrics.hpp>
+  # #include <overlay/messages.hpp>
```
asekretenko: Can this be fixed in "agent.hpp"?

overlay/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+  ${CMAKE_CURRENT_SOURCE_DIR}/..
+)
+
+add_library(overlay SHARED
+  agent_metrics.cpp
+  agent.cpp
+  # LNK2019: unresolved external symbol "public: __cdecl mesos:: log::Log::Log
+  # `log` is explicitly excluded for Windows in mesos
+  # master_metrics.cpp
+  # master.cpp
```
asekretenko: Took me some time out that this comment refers to Mesos replicated log that is not supported on Windows; could you rephrase this comment to make this more clear?

cmake/Dependencies.cmake
```diff
@@ -0,0 +1,247 @@
...
+  PATHS ${MESOS_ROOT}/build/src
+  PATH_SUFFIXES Release
+)
+find_library(
+  MESOS_LIB_DEBUG
+  NAMES mesos
+  PATHS ${MESOS_ROOT}/build/src
+  PATH_SUFFIXES Debug
+)
+add_library(mesos STATIC IMPORTED)
```
asekretenko: Do I understand correctly that this statically links Mesos into each dll? 
If yes, then what is the resulting size of each dll?

CMakeLists.txt
```diff
@@ -0,0 +1,198 @@
+
+if (NOT WIN32)
+  cmake_minimum_required(VERSION 3.7.0)
+else ()
+  cmake_minimum_required(VERSION 3.8.0)
+endif ()
+
+project(dcos-mesos-modules)
+
+get_filename_component(MESOS_ROOT ../mesos ABSOLUTE)
```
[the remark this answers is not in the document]
akornatskyy: Yes, done.

CMakeLists.txt
```diff
@@ -0,0 +1,198 @@
...
+include(Dependencies)
+
+# Set the default standard to C++11 for all targets.
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+# Do not use, for example, `-std=gnu++11`.
+set(CMAKE_CXX_EXTENSIONS OFF)
+
+if (CMAKE_CXX_COMPILER_ID MATCHES MSVC)
+  # TODO(andschwa): Switch to `/W4` and re-enable possible-loss-of-data warnings.
```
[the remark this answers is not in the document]
akornatskyy: done

CMakeLists.txt
```diff
@@ -0,0 +1,198 @@
...
+    /W3
+    # Disable permissiveness.
+    /permissive-
+    # C4244 is a possible loss of data warning for integer conversions.
+    /w44244
+    # C4267 is a possible loss of data warning when converting from `size_t`.
+    /w44267)
+endif ()
+
+# WINDOWS CONFIGURATION.
```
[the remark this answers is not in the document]
akornatskyy: done

HOWTO.md
```diff
@@ -0,0 +1,10 @@
+# CMAKE
```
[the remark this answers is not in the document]
akornatskyy: removed.

CMakeLists.txt
akornatskyy: That is not kind of TODO due to some limitation stated. This cannot be fixed immediately, without touching a larger refactoring field.

metrics/isolator.hpp
akornatskyy: The reason of `extern "C"` is to have the specific name visible and as such have an ability to load module.

overlay/CMakeLists.txt
akornatskyy: No. That will break linux build.

overlay/CMakeLists.txt
akornatskyy: No. The above is clear to me. If you have a better way to make this more clear, give me some text and I will do replacement to avoid ping-pong.

cmake/Dependencies.cmake
akornatskyy: Yes.

libjournaldlogger.dll 5.0M
logsink.dll 267K
metrics.dll 5.3M

CMakeLists.txt
akornatskyy: https://github.com/dcos/dcos-mesos-modules/blob/master/Makefile.am#L192-L198

overlay/agent.hpp
```diff
@@ -13,6 +13,13 @@
 #include <overlay/agent_metrics.hpp>
 #include <overlay/messages.hpp>
 
+// `REGISTERING` is used as an enum value, but it's actually defined as a
+// constant in the Windows SDK.
+#ifdef __WINDOWS__
+#undef REGISTERING
+#undef REGISTERED
+#endif // __WINDOWS__
+
```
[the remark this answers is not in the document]
asekretenko: Do I understand correctly that the conflicting definitions just come from this file (ManagerProcess::State below), and are not #include`d from some Mesos headers? 

If so, can you rename them now or, at least,  leave `//TODO(akornatskyy): Rename enum values in ManagerProcess::State in this file to avoid name collistion.` ?

overlay/agent.cpp
```diff
@@ -558,7 +559,7 @@ void ManagerProcess::doReliableRegistration(Duration maxBackoff)
 
   // Determine the delay for next attempt by picking a random
   // duration between 0 and 'maxBackoff'.
-  Duration backoff = maxBackoff * ((double) ::random() / RAND_MAX);
+  Duration backoff = maxBackoff * ((double) rand() / RAND_MAX);
```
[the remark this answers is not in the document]
asekretenko: Could you please either convert these to `std::uniform_real_distribution<double>` now or leave in the code a `//TODO` comment to do this in the future?

overlay/agent.hpp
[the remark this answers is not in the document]
akornatskyy: added TODO

overlay/agent.cpp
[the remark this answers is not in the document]
akornatskyy: converted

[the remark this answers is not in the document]
akornatskyy: I will revert my change since it appears to be compiling on windows but breaks build for linux.

overlay/CMakeLists.txt
asekretenko: Found https://issues.apache.org/jira/browse/MESOS-5932 for porting replicated log to Windows (need to update leveldb first).

I would suggest, instead of putting error message into the comment, to point to the real underlying issue, like
```
# TODO(akornatskyy): Windows build of overlay module master (which depends on Mesos replicated log)
# can be re-enabled after replicated log is available on Windows (see MESOS-5932).
```

CMakeLists.txt
```diff
@@ -0,0 +1,199 @@
...
+add_subdirectory(logsink)
+
+# FIXME(akornatskyy): due to import path in 'overlay/messages.proto'
+# NOTE: according to [1], overlay module is not required for Windows
+# [1] https://github.com/mesosphere/dcos-enterprise/tree/master/packages/mesos/extra/windows-ee.patches
+#add_subdirectory(overlay)
+
+
+# FIXME(akornatskyy): due to limitation of protobuf generation process
+# and include in source
```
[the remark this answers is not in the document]
akornatskyy: No, it is not... this has no relation to cmake specifics, the issue is with source code layout and relative paths which incorrectly resolve. I have a task in backlog so that is not an year to get fixed. As part of port to cmake linux, it would require refactor the code a bit to fix.

CMakeLists.txt
akornatskyy: done

metrics/isolator.hpp
akornatskyy: right

CMakeLists.txt
akornatskyy: done

overlay/CMakeLists.txt
akornatskyy: done

PR #133, opened 2020-03-18
Ported to cmake on linux.

akornatskyy:


CMakeLists.txt
```diff
@@ -0,0 +1,217 @@
...
+    "${MESOS_ROOT}/build/src"
+    "${MESOS_ROOT}/include"
+    "${MESOS_ROOT}/src"
+    "${NVML_INCL}"
+    "${PICOJSON_INCL}"
+    "${PROCESS_INCL}"
+    "${PROTOBUF_INCL}"
+    "${RAPIDJSON_INCL}"
+    "${STOUT_INCL}"
+    "${ZOOKEEPER_INCL}"
```
asekretenko, line 153: Can you compose the list of include dirs for tests once and reuse it for all the test binaries?
like
```
list(APPEND TEST_INCLUDE_DIRS ${MESOS_ROOT}/include ${MESOS_ROOT}/src ...)
...
target_include_directories(test-metrics PRIVATE ${TEST_INCLUDE_DIRS})
```

The same applies to lists of libraries in `target_link_libraries()`

cmake/Dependencies.cmake
```diff
@@ -0,0 +1,99 @@
...
+  "${MESOS_ROOT}/3rdparty/stout/include")
+set (ZOOKEEPER_INCL
+  "${MESOS_ROOT}/include/zookeeper"
+  "${MESOS_ROOT}/build/3rdparty/zookeeper-${ZOOKEEPER_VERSION}/src/zookeeper-${ZOOKEEPER_VERSION}/src/c/generated"
+  "${MESOS_ROOT}/build/3rdparty/zookeeper-${ZOOKEEPER_VERSION}/src/zookeeper-${ZOOKEEPER_VERSION}/src/c/include")
+
+set (GLOG_INCL
+  "${MESOS_ROOT}/include"
+  "${MESOS_ROOT}/build/3rdparty/glog-${GLOG_VERSION}/src/glog-${GLOG_VERSION}-install/include")
+find_library(
```
asekretenko, line 26: If we end up relying on `-DENABLE_INSTALL_MODULE_DEPENDENCIES` (I think we should), then we should add `NO_DEFAULT_PATH` in all invocations of find_library()  for bundled libraries.

Rationale: developing mesos modules is much simpler when you don't have to think: "Damn, I have glog installed on my system, which glog were the modules built against? was it the bundled one or the system one?..."

CMakeLists.txt
```diff
@@ -0,0 +1,217 @@
...
+
+set(MESOS_ROOT "../mesos" CACHE PATH "Mesos root directory.")
+get_filename_component(MESOS_ROOT ${MESOS_ROOT} ABSOLUTE)
+set(BOOST_ROOT_DIR "" CACHE PATH "Boost libs root directory.")
+
+message("MESOS_ROOT is ${MESOS_ROOT}")
+
+# CMAKE MODULE SETUP.
+#####################
+# Paths that are searched when `include(...)` is called.
```
asekretenko, line 13: I'm a bit lost: among the three `include(...)`s I see in this patch
```
include(CTest)
include(Versions OPTIONAL)
include(Dependencies)
``` 
which one is included from ${MESOS_ROOT}/...?

cmake/Dependencies.cmake
```diff
@@ -0,0 +1,99 @@
...
+
+if ("${BOOST_ROOT_DIR}" STREQUAL "")
+  set (BOOST_INCL "${MESOS_ROOT}/build/3rdparty/boost-${BOOST_VERSION}/src/boost-${BOOST_VERSION}")
+else ()
+  set (BOOST_INCL ${BOOST_ROOT_DIR}/include)
+endif ()
+
+set (PICOJSON_INCL
+  "${MESOS_ROOT}/include"
+  "${MESOS_ROOT}/build/3rdparty/picojson-${PICOJSON_VERSION}/src/picojson-${PICOJSON_VERSION}")
```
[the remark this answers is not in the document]
greggomann: Good call @asekretenko , I like the idea of just assuming that Mesos has been build with ENABLE_INSTALL_MODULE_DEPENDENCIES.

README.md
```diff
@@ -40,9 +40,12 @@ should be added to the `configure`'s command-line arguments too.
 cd <mesos-source>
 mkdir build
 cd build
-cmake .. -T "host=x64" -DENABLE_SSL=ON -DCMAKE_BUILD_TYPE=Release
-# cmake .. -T "host=x64" -DENABLE_SSL=ON -DBUILD_TESTING=OFF
+# windows
+cmake .. -T "host=x64" -DENABLE_SSL=ON -DBUILD_TESTING=OFF -DCMAKE_BUILD_TYPE=Release
```
greggomann, line 44: Is this still relevant? And below.

CMakeLists.txt
```diff
@@ -0,0 +1,217 @@
...
+
+# Set the default standard to C++11 for all targets.
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+# Do not use, for example, `-std=gnu++11`.
+set(CMAKE_CXX_EXTENSIONS OFF)
+
+add_subdirectory(dockercfg)
+add_subdirectory(logsink)
+add_subdirectory(journald)
```
greggomann, line 32: Why do you opt for having separate `CMakeLists.txt` files in these module subdirectories, while others like overlay and metrics have their code inlined in this file?

cmake/Dependencies.cmake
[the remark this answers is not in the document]
akornatskyy: > Is it true that all these include dirs from build tree (and also build tree dirs in find_library()/find_program()) are needed to be able to build modules without installed module dependencies?

yes. this way we point into a fresh build of mesos. if we do require module developer to use `ENABLE_INSTALL_MODULE_DEPENDENCIES` option that would make module's cmake file dependencies much cleaner, plus an extra step to install mesos somewhere...

are we good to proceed with this change?

tests/metrics_tests.cpp
```diff
@@ -111,7 +111,7 @@ class MetricsTest : public MesosTest
         ip + ":" + port,
         "/" + METRICS_PROCESS + "/" + API_PATH,
         stringify(REQUEST_TIMEOUT));
-    
+
```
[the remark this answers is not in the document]
akornatskyy: no, extra space are automatically deleted by editor.

CMakeLists.txt
akornatskyy: yep. I will use the approach in `dcos-ee-mesos-modules`, with a list of includes.

cmake/Dependencies.cmake
akornatskyy: right... awaiting votes/decision about this topic in other PR.

CMakeLists.txt
akornatskyy: `Versions`; it will not be necessary if we compile with an installed version of mesos. on hold for now until decision whenever require mesos install for module development.

README.md
akornatskyy: no, will clean up.

CMakeLists.txt
akornatskyy: I would prefer to have `CMakeLists.txt` in each directory (per module), however, right now that is not possible because corresponding changes require some changes to source code (that would break autotools build). Probably it makes sense to either add a TODO here or refactor modules layout of a part of autotools removal. thoughts?

cmake/Dependencies.cmake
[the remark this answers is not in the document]
asekretenko: Let's proceed with this change.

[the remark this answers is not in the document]
akornatskyy: done, updated dcos ref and tested. it works. please note, I have added TODO for tests (in `Dependencies.cmake`). The point is that mesos install should also provide tests modules (and also probably gmock/gtest). I believe we agreed that to be out of this PR.

CMakeLists.txt
asekretenko: Please add a TODO here, so that after we remove autotools we don't forget that it was the thing that caused this layout.

CMakeLists.txt
asekretenko: Now that we settled on using -DENABLE_INSTALL_MODULE_DEPENDENCIES, can `Versions` be just dropped?

cmake/Dependencies.cmake
asekretenko: We settled on using -DENABLE_INSTALL_MODULE_DEPENDENCIES, this means that we can add `NO_DEFAULT_PATH`, right?

CMakeLists.txt
akornatskyy: added.

CMakeLists.txt
akornatskyy: yes, removed.

cmake/Dependencies.cmake
akornatskyy: yes, added `NO_DEFAULT_PATH` to each of `find_library` in `Deendencies.cmake`.

PR #134, opened 2020-03-25
Fixed test cases for windows.

akornatskyy:
https://jira.d2iq.com/browse/D2IQ-65487

### DIFF
--- a/journald/modules.json.cmake-in
+++ b/journald/modules.json.cmake-in
@@ -1,0 +1,30 @@
+{
+  "libraries": [
+    {
+      "file": "@CMAKE_CURRENT_BINARY_DIR@/@CMAKE_BUILD_TYPE@/libjournaldlogger.dll",
+      "modules": [
+        {
+          "name": "com_mesosphere_mesos_JournaldLogger",
+          "parameters": [
+            {
+              "key": "companion_dir",
+              "value": "@CMAKE_CURRENT_BINARY_DIR@/@CMAKE_BUILD_TYPE@/"
+            }, {
+              "key": "libprocess_num_worker_threads",
+              "value": "2"
+            }, {
+              "key": "destination_type",
+              "value": "journald+logrotate"
+            }, {
+              "key": "fluentbit_ip",
+              "value": "127.0.0.1"
+            }, {
+              "key": "fluentbit_port",
+              "value": "5170"
+            }
+          ]
+        }
+      ]
+    }
+  ]
+}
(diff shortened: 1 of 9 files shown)


DominikDary:
@asekretenko please review

journald/modules.json.cmake-in
```diff
@@ -0,0 +1,30 @@
+{
+  "libraries": [
+    {
+      "file": "@CMAKE_CURRENT_BINARY_DIR@/@CMAKE_BUILD_TYPE@/libjournaldlogger.dll",
```
asekretenko: How does one configure and build modules so that `CMAKE_BUILD_TYPE` is substituted correctly?

(I tried to perform clean build with cmake as described in README.md - looks like `CMAKE_BUILD_TYPE` is empty at the time when substitution is performed).

tests/journald_tests.cpp
```diff
@@ -158,6 +158,7 @@ class JournaldLoggerTest : public MesosTest,
     modules = _modules.get();
 
     // Initialize the modules.
+    // TODO(akornatskyy): The following line crashes in Windows
```
asekretenko: Is this still true? 
In my clean build, the only problem seems so be the mishandled substitution of `CMAKE_BUILD_TYPE` (see above); if I type that in manually into the `modules.json`, this line seems to work.

tests/journald_tests.cpp
```diff
@@ -312,7 +313,7 @@ TEST_F(JournaldLoggerTest, ROOT_LogToJournaldWithBigLabel)
   AWAIT_READY(secondQuery);
   ASSERT_FALSE(strings::contains(secondQuery.get(), specialString));
 }
-#endif // __WINDOWS__
+#endif // __linux__
```
asekretenko: Please either add  a matching `#ifdef`, or remove this (depending on whether the issue with CMAKE_BUILD_TYPE exists/remains or not).

journald/modules.json.cmake-in
akornatskyy: `DCMAKE_BUILD_TYPE` needs to be passed during configuration (updated README to reflect this).

tests/journald_tests.cpp
akornatskyy: no, removed

tests/journald_tests.cpp
akornatskyy: the matching `#ifdef` starts at line 184; that used to be misprint that I fixed.

DominikDary:
@greggomann or @bmahler have you permissions to land this?

PR #135, opened 2020-05-12
Remove the current cmake build system.

akornatskyy:
## High-level description

This PR removes the current cmake build system, which only works on Windows, in preparation for the addition of a new Linux-specific cmake build system.

## Changelog automation

[DCOS_OSS-68400](https://jira.d2iq.com/browse/D2IQ-68400)


greggomann:
Rather than removing the Windows cmake build, it looks like this PR removes the cmake build entirely from dcos-mesos-modules? Is that really what we want to do?

akornatskyy:
> Rather than removing the Windows cmake build, it looks like this PR removes the cmake build entirely from dcos-mesos-modules? Is that really what we want to do?

Yes, that is specifically what I am aiming. The reason is that master right now have merged windows version. If in the other PR I will remove windows and add linux that will be very confusing due to way of too many changes (including formatting issues due to removed `if` statements, etc).

Thus you either merge this PR and see only linux related cmake PR or close this one and likely will have a difficulty in review of linux PR. up to you.

